PR #113, opened 2014-12-11
compile with libuv 1.x

czras:
Error handling is not complete, as it needs some more changes, but it is functional and compiling.


mpenick:
Started with this PR, updated error handling and create a new PR: https://github.com/datastax/cpp-driver/pull/123. Up for review.

PR #120, opened 2015-01-07
Test correct flag for using external boost

richardkchapman:
When including the cmake file from a parent cmake project, Boost_FOUND may
be set even though we don't want to use an external boost (for example,
because the version of boost is not new enough to support the test code)


mpenick:
Thanks, good catch.

PR #121, opened 2015-01-07
Use memcpy to avoid strict aliasing warnings with gcc 4.4

ghalliday:
We hit this problem compiling on Ubuntu 10.04 with g++ 4.4.

This is a better solution to avoid the strict aliasing warnings when casting
a float/double to an integer.  Strangely, older versions of gcc are better at
reporting a warning in this situation than more recent versions.

The previous copy_cast solution to the alias problem doesn't really solve the problem.  The
code is still accessing a float variable through an integer pointer.  The memcpy of the pointer
inside the copy_cast is enough to stop the compiler tracking that it is an alias, but interestingly
on 4.4 the compiler does track it - hence the warning.

This change uses memcpy() to copy the value, so the same memory location is never aliased.

I compiled the code and as far as I could see from the assembler output the generated code was the same (g++ 4.8.2) before after after the patch.

Signed-off-by: Gavin Halliday <gavin.halliday@lexisnexis.com>

### DIFF
--- a/src/serialization.hpp
+++ b/src/serialization.hpp
@@ -116,22 +116,32 @@ inline char* decode_int64(char* input, cass_int64_t& output) {
 
 inline void encode_float(char* output, float value) {
   BOOST_STATIC_ASSERT(std::numeric_limits<float>::is_iec559);
-  encode_int32(output, *copy_cast<float*, int32_t*>(&value));
+  int32_t int_value;
+  memcpy(&int_value, &value, sizeof(int_value));
+  encode_int32(output, int_value);
 }
 
 inline char* decode_float(char* input, float& output) {
   BOOST_STATIC_ASSERT(std::numeric_limits<float>::is_iec559);
-  return decode_int32(input, *copy_cast<float*, int32_t*>(&output));
+  int32_t int_value;
+  char* pos = decode_int32(input, int_value);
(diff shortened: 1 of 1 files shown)


ghalliday:
This is the second version - the first version is on my strictalias branch, and uses a union, but that isn't strictly conforming.  This one is probably cleaner.

I only changed these examples that use copy_cast, although I think most (all) of the other instances could be replaced with reinterpret_cast:

ssll_open_sll_imp.cpp(253) - this is ok because it is legal to add switch between signed/unsigned variants.
the socket casts are ok because it is casting to the real type of the object, similarly with the uv casts.

if Md5::body used a unsigned char instead of a uint8_t then that cast would also be ok.

Please get back to me if anything needs modifying.

mpenick:
Thanks. Good call on the unnecessary uses of copy_cast<>. 

I've seen some code add an extra indirection, by casting to uintptr_t, to get around that gcc 4.4 issue.  Here's an example, https://code.google.com/p/v8/source/browse/branches/bleeding_edge/src/utils.h?r=18578#942.

mpenick:
I merged your changes and remove copy_cast<>. Just testing on a couple different compilers, including gcc-4.4. I'll probably get this change in today.

PR #123, opened 2015-01-15
CPP-195 - Finished support for libuv v1.x

mpenick:


src/address.hpp
```diff
@@ -42,6 +42,10 @@ class Address {
 
   bool init(const struct sockaddr* addr);
 
+  const struct sockaddr* addr() const {
+    return copy_cast<const struct sockaddr_storage*, const struct sockaddr*>(&addr_);
+  }
+
```
aholmberg: Curious why this was added.

aholmberg: Never mind. IDE search fail was not showing usage.

src/loop_thread.hpp
```diff
@@ -26,7 +26,6 @@ class LoopThread {
 public:
   LoopThread()
       : loop_(uv_loop_new())
```
aholmberg: It  looks like we're still using the deprecated loop functions here. Did you intend to change that for libuv 1.x to address [CPP-194](https://datastax-oss.atlassian.net/browse/CPP-194)?

mpenick: That's fixed in this PR: https://github.com/datastax/cpp-driver/pull/126

aholmberg: Ok. You never mentioned that PR in CPP-194, so I thought this was the only one.

PR #124, opened 2015-01-15
CPP-197 - Reduced the default number of threads and connections

mpenick:


src/cluster.cpp
```diff
@@ -56,8 +56,8 @@ CassError cass_cluster_set_protocol_version(CassCluster* cluster,
 
 void cass_cluster_set_num_threads_io(CassCluster* cluster,
                                      unsigned num_threads) {
-  // 0 is okay, it means use a thread per core
-  cluster->config().set_thread_count_io(num_threads);
+  // At least one thread is required
+  cluster->config().set_thread_count_io(num_threads > 0 ? num_threads : 1);
```
aholmberg: Should this return an error when ``num_threads == 0``, as in some of the other set config functions?
(e.g. ``cass_cluster_set_queue_size_io``)

mpenick: You're right that's more consistent.

PR #125, opened 2015-01-16
join() call is not happening in release builds

richardkchapman:
The assert() macro expands to do nothing in a release build, so any side
effects of the expression being asserted do not occur.

This was causing random crashes and memory corruptions in a release-mode build
of the driver.

### DIFF
--- a/src/loop_thread.hpp
+++ b/src/loop_thread.hpp
@@ -59,7 +59,9 @@ public:
   void join() {
     if (is_joinable_) {
       is_joinable_ = false;
-      assert(uv_thread_join(&thread_) == 0);
+      int rc = uv_thread_join(&thread_);
+      (void) rc;
+      assert(rc == 0);
     }
   }
 


sandvige:
You should remove the if statement no?

COLE Edouard,
RG System

Le 16 janv. 2015 à 12:41, Richard Chapman <notifications@github.com<mailto:notifications@github.com>> a écrit :


The assert() macro expands to do nothing in a release build, so any side
effects of the expression being asserted do not occur.

This was causing random crashes and memory corruptions in a release-mode build
of the driver.

Signed-off-by: Richard Chapman rchapman@hpccsystems.com<mailto:rchapman@hpccsystems.com>

________________________________
You can merge this Pull Request by running

  git pull https://github.com/hpcc-systems/cpp-driver assert-join

Or view, comment on, or merge it at:

  https://github.com/datastax/cpp-driver/pull/125

Commit Summary

  *   join() call is not happening in release builds

File Changes

  *   M src/loop_thread.hpp<https://github.com/datastax/cpp-driver/pull/125/files#diff-0> (4)

Patch Links:

  *   https://github.com/datastax/cpp-driver/pull/125.patch
  *   https://github.com/datastax/cpp-driver/pull/125.diff

richardkchapman:
the if() statement is to avoid errors about an unused local variable.

sandvige:
You should have wrote something like that:
``` cpp
int rc = uv_thread_join(&thread_);
(void)rc;
assert(rc == 0);
```

richardkchapman:
That's certainly clearer, and works on g++ to suppress the warning. I _think_ I've seen it provoke other warnings (expression has no effect, or similar) on other compilers in the past, which is why I tend not to use it. But if it's the preferred style on this project I'm happy to change it.

sandvige:
Oh no, I don't know if it's the preferred way here, I'm just saying as it's the way most cross platform framework do, like Qt with Q_UNUSED() macro. It's also discussed here:
- http://stackoverflow.com/questions/4030959/will-a-variablename-c-statement-be-a-no-op-at-all-times/4030983#4030983
- http://stackoverflow.com/questions/19576884/does-q-unused-have-any-side-effects

richardkchapman:
Well, I changed it. It might give a warning on some old MS compilers.

mpenick:
Thanks. I made sure there weren't any other asserts with necessary side effect. :S

PR #136, opened 2015-02-16
Fix compile error on Clang 6.0

richardkchapman:
Fix build failure compiling on OSX (clang reports version Apple LLVM
version 6.0 (clang-600.0.56))

### DIFF
--- a/src/logger.cpp
+++ b/src/logger.cpp
@@ -82,7 +82,10 @@ void Logger::LogThread::log(CassLogLevel severity,
     ""
   };
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
   vsnprintf(message.message, sizeof(message.message), format, args);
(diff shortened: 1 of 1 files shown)


mpenick:
I'm unable to reproduce the warning on:
```
Apple LLVM version 6.0 (clang-600.0.56) (based on LLVM 3.5svn)
```

I even tried explicitly enabling that warning. Maybe it's not an issue on latest 1.0? The issue could be fixed by propagating the printf attribute information:

https://github.com/mpenick/cpp-driver/commit/87fa1836e81992022f89a2712d8962dca9fffa17

Does that fix the issue for you?

richardkchapman:
Yes, that does seem to fix the issue (and it's a better way to fix it than my change) - thanks.

mpenick:
Pushed to master https://github.com/datastax/cpp-driver/commit/200bc5cb9bbcaa544050a94c511a0c8aa69c97c2

PR #137, opened 2015-02-17
CPP-209 - Lowered Boost version to 1.54

mpenick:
Boost 1.54 compiles and runs the unit/integration tests.

Improved error output for versions of Boost older than 1.54 or when Boost
is not found.


mikefero:
+1

Updated how the version information was displayed on error

PR #138, opened 2015-02-17
CPP-160 - Adding Testing Documentation, Updating Building Documentation  and Enhancing/Refactoring Build Configuration

mikefero:
Copyright information was also updated


examples/ssl/ssl.c
```diff
@@ -45,14 +46,16 @@ int load_trusted_cert_file(const char* file, CassSsl* ssl) {
...
-  fread(cert, sizeof(char), cert_size, in);
+  bytes_read = fread(cert, sizeof(char), cert_size, in);
   fclose(in);
 
-  rc = cass_ssl_add_trusted_cert(ssl, cass_string_init2(cert, cert_size));
-  if (rc != CASS_OK) {
-    fprintf(stderr, "Error loading SSL certificate: %s\n", cass_error_desc(rc));
-    free(cert);
-    return 0;
+  if (bytes_read == (sizeof(char) * cert_size)) {
```
mpenick: sizeof(char) unnecessary as it is always 1.

test/unit_tests/src/test_async_queue.cpp
```diff
@@ -37,9 +36,20 @@ struct TestAsyncQueue : public cass::LoopThread {
   TestAsyncQueue(size_t queue_size)
     : value_(0)
     , async_queue_(queue_size) {
+    init();
     async_queue_.init(loop(), this, TestAsyncQueue::async_func);
```
mpenick: We should probably assert these two `init()` calls.

examples/ssl/ssl.c
mikefero: agree ... updated instances of sizeof(char) to 1 where applicable

test/unit_tests/src/test_async_queue.cpp
mikefero: +1 on changes

PR #146, opened 2015-03-19
CPP-187 - Enhance Integration Tests for Cassandra v1.2.x

mikefero:
Integration tests that could run against C* v1.2.x have been fixed along with the skipping of batch and serial_consistency suites.


mpenick:
@mikefero I would prefer is we only use prepared statements if the version <= 1.2, otherwise we should use a bound statements.

mikefero:
@mpenick Addressed comments and brought inline with the 2.1.0-beta by updating to master and cherry picking commits for v3 protocol in CPP-260/PR-170. Force committed to remove previous version this PR to ensure no merge conflict resolution would be needed (manual merged previous changes and addressed API changes where needed)

mpenick:
Merged in 20d457c295f208a15f9ddba425f7e2ea00d74329

PR #147, opened 2015-03-19
CPP-242 - Added string functions for the CassInet type

mpenick:


src/types.cpp
```diff
@@ -56,6 +58,32 @@ CassInet cass_inet_init_v6(const cass_uint8_t* address) {
...
 
+void cass_inet_string(CassInet inet, char* output) {
+  uv_inet_ntop(inet.address_length == CASS_INET_V4_LENGTH ? AF_INET : AF_INET6,
+               inet.address,
+               output, CASS_INET_STRING_LENGTH);
+}
+
+CassError cass_inet_from_string(const char* str, CassInet* output) {
+#if UV_VERSION_MAJOR == 0
+  if (uv_inet_pton(AF_INET, str, output->address).code == UV_OK) {
```
avalanche123: this makes it impossible to create an ipv6 inet value on a system that doesn't have ipv6 support right?

mpenick: @avalanche123 libuv implements custom implementations of those functions. https://github.com/libuv/libuv/blob/v1.x/src/inet.c#L161-L313

include/cassandra.h
```diff
@@ -131,6 +131,11 @@ typedef struct CassString_ {
 #define CASS_INET_V6_LENGTH 16
 
 /**
+ * The size of an inet string including a null terminator.
+ */
+#define CASS_INET_STRING_LENGTH 46
```
aholmberg: No change requested, but just an aside:
I kind of liked the way they defined this size with no 'magic numbers' in the old branch of libuv:
https://github.com/libuv/libuv/blob/v0.10/src/inet.c#L81

It's removed in 1.0, though.

aholmberg:
+1 on changes

include/cassandra.h
mpenick: That's pretty cool. I'm going to keep the magic number because it follows the pattern of the other magic constants.

mpenick:
Final commit: 39aa618b2a34842e02af8ca3510ae9ab427708f3

PR #153, opened 2015-04-03
CPP-150 - integration test for Latency Aware Policy

mikefero:
This test will ensure latency is incurred on all hosts connected in a
session. A new feature in CCM bridge is being used to send signals
(SIGSTOP/SIGCONT) on POSIX platform or pausing/resuming all threads in a
process on Windows. Since this is not deterministic an acceptable
threshold is used for latencies that are less than expected.

This commit also enhances host logging of latency and allow for Boost
location to be specified on the command line through BOOST_ROOT_DIR.

### DIFF
--- a/test/integration_tests/src/test_latency_aware_policy.cpp
+++ b/test/integration_tests/src/test_latency_aware_policy.cpp
@@ -1,0 +1,245 @@
+/*
+  Copyright (c) 2014-2015 DataStax
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+#ifdef STAND_ALONE
+#   define BOOST_TEST_MODULE cassandra
+#endif
+
+#include <boost/test/unit_test.hpp>
+#include <boost/test/debug.hpp>
+#include <boost/thread.hpp> // Sleep functionality
+
+#include <uv.h>
+
+#include "cassandra.h"
+#include "testing.hpp"
+#include "test_utils.hpp"
+
+#define DEFAULT_CASSANDRA_NODE_PORT 9042
+
+struct LatencyAwarePolicyTest {
+public:
+  boost::shared_ptr<cql::cql_ccm_bridge_t> ccm_;
+
+  LatencyAwarePolicyTest()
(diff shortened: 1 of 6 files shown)


test/integration_tests/src/test_latency_aware_policy.cpp
```diff
@@ -0,0 +1,245 @@
...
+      CassError error_code = test_utils::wait_and_return_error(future.get(),
+                                                               240 * test_utils::ONE_SECOND_IN_MICROS);
+
+      // Ignore all timing errors
+      if (error_code != CASS_OK &&
+          error_code != CASS_ERROR_LIB_REQUEST_TIMED_OUT &&
+          error_code != CASS_ERROR_SERVER_READ_TIMEOUT) {
+        CassString message;
+        cass_future_error_message(future.get(), &message.data, &message.length);
+        BOOST_FAIL(std::string(message.data, message.length) << "' (" << cass_error_desc(error_code) << ")");
```
mpenick: I'm weary of calling into Boost test from a thread that's not the main thread. It might be ok in this case because it's the only thread, but i'm not sure. It might have problems if this fails (throws). An alternative solution would be to print to stderr,  mark a failed flag and break the loop. The failed flag can be checked in the main thread.

test/integration_tests/src/test_latency_aware_policy.cpp
```diff
@@ -0,0 +1,245 @@
...
+      }
+  }
+
+private:
+  static test_utils::CassSessionPtr session_;
+  const cql::cql_ccm_bridge_configuration_t& configuration_;
+  test_utils::CassClusterPtr cluster_;
+  static bool is_running_;
+#if UV_VERSION_MAJOR == 0
+  static uv_loop_t* loop_;
```
mpenick: I can't find an use of this loop. Is this necessary?

mikefero: loop is used after the single thread is created in start_query_execution in start_thread here; https://github.com/datastax/cpp-driver/pull/153/files#diff-20b2a3bbb7595403a32ccb337f67af91R139. Is this implemented wrong?

test/integration_tests/src/test_latency_aware_policy.cpp
mikefero: Good point.

test/integration_tests/src/test_latency_aware_policy.cpp
```diff
@@ -0,0 +1,245 @@
...
+ * to simulate network latency on a node
+ *
+ * @since 2.0.0
+ * @jira_ticket CPP-150
+ * @test_category load_balancing_policy:latency_aware
+ */
+BOOST_AUTO_TEST_CASE(ensure_latency) {
+  // Start gathering latencies for nodes
+  start_query_execution();
+  boost::this_thread::sleep_for(boost::chrono::milliseconds(1000));
```
mpenick: This should use a condition variable to wait.

mpenick:
Final commit: de966a30baa372def8cabd4279e848bc54e85883

PR #159, opened 2015-05-05
Pick the right STRERROR_R macro for FreeBSD

lehenbauer:
FreeBSD needs the same STRERROR_R that OS X does.


lehenbauer:
This pull request is in error.  I will resubmit.

PR #160, opened 2015-05-05
Pick the correct STRERROR_R macro for FreeBSD

lehenbauer:
FreeBSD needs the same STRERROR_R definition used for OS X.

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -61,7 +61,7 @@ namespace cass {
 
 #define STRERROR_BUFSIZE_ 256
 
-#if defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


mpenick:
Thanks. I will make sure to  keep FreeBSD in mind moving forward.

lehenbauer:
Thanks, Michael... much appreciated!  There's probably some cleaner way to use cmake to detect which version of STRERROR_R to use.  Google shed a little light on the subject.

mpenick:
@lehenbauer you're absolutely right about that doing this in cmake would be much better.

PR #161, opened 2015-05-07
CPP-256/251 - Removed logging thread (1.0.2)

mpenick:


include/cassandra.h
```diff
@@ -20,7 +20,7 @@
 #include <stddef.h>
 
 #if !defined(CASS_STATIC)
-#  if (defined(WIN32) || defined(_WIN32))
+#  if defined(_MSC_VER)
```
mikefero: In the event MinGW is used to compile the source the symbol will not be exported to the DLL. Was that the intention?

PR #162, opened 2015-05-07
CPP-256/251 - Removed logging thread (2.0.1)

mpenick:


include/cassandra.h
```diff
@@ -20,7 +20,7 @@
 #include <stddef.h>
 
 #if !defined(CASS_STATIC)
-#  if defined(_WIN32)
+#  if defined(_MSC_VER)
```
mikefero: Same from PR 161:

In the event MinGW is used to compile the source the symbol will not be exported to the DLL. Was that the intention?

mpenick: Good call. Thanks for reviewing.

PR #163, opened 2015-05-07
CPP-257 - Fixed crash/hang when 'system.local' is empty (2.0.1)

mpenick:


aholmberg:
Looks good. Does the test from #164 need to be added here, or will that get merged at another time?

PR #170, opened 2015-06-01
CPP-260 - Added initial support for protocol v3

mpenick:


src/data_type.hpp
```diff
@@ -0,0 +1,449 @@
...
+        if(!types_[i]->equals(collection_type->types_[i], true)) {
+          return false;
+        }
+      }
+    } else if(!types_.empty() && !collection_type->types_.empty()) { // An empty collection doesn't have subtypes
+      if (types_.size() != collection_type->types_.size()) {
+        return false;
+      }
+      for (size_t i = 0; i < types_.size(); ++i) {
+        if(!types_[i]->equals(collection_type->types_[i], false)) {
```
aholmberg: I think if-exact-else could be simplified removing the conditional and passing `exact` instead of the bool literal.

src/batch_request.cpp
```diff
@@ -17,14 +17,14 @@
...
+#include "external_types.hpp"
 #include "serialization.hpp"
 #include "statement.hpp"
-#include "types.hpp"
 
 
 extern "C" {
 
-CassBatch* cass_batch_new(CassBatchType type) {
+CassBatch* cass_batch_new(CassSession* session, CassBatchType type) {
```
aholmberg: Is session just added here to future-proof and align with cass_collection_new?

src/statement.cpp
```diff
@@ -16,160 +16,27 @@
...
-CassStatement* cass_statement_new(const char* query, size_t parameter_count) {
-  return cass_statement_new_n(query, strlen(query), parameter_count);
+CassStatement* cass_statement_new(CassSession* session,
+                                  const char* query,
+                                  size_t parameter_count) {
+  return cass_statement_new_n(session, query, strlen(query), parameter_count);
 }
 
-CassStatement* cass_statement_new_n(const char* query,
+CassStatement* cass_statement_new_n(CassSession* session,
```
aholmberg: Same question as in cass_batch_new

include/cassandra.h
```diff
@@ -2942,6 +2955,11 @@ cass_collection_append_decimal(CassCollection* collection,
                                size_t varint_size,
                                cass_int32_t scale);
 
+
+CASS_EXPORT CassError
+cass_collection_append_collection(CassCollection* collection,
```
aholmberg: Seem to be missing a definition for this. Already talked about it. I assume it will be added with examples, when needed.

src/hash_index.cpp
```diff
@@ -0,0 +1,115 @@
+/*
```
aholmberg: Nice generalization of this metadata stuff.

src/collection.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include "macros.hpp"
+
+#include <string.h>
+
+extern "C" {
+
+CassCollection* cass_collection_new(CassSession* session,
+                                    CassCollectionType type,
+                                    size_t item_count) {
+  return CassCollection::to(new cass::Collection(session->protocol_version(),
```
aholmberg: One concern about instantiating/encoding the collection with the session protocol version: if we're connected to a mixed-version cluster, the CC may negotiate to one version, and the connection this ends up on may have negotiated a lower version.

include/cassandra.h
aholmberg: One thing to consider in this implementation: in v2, nested inner collections are encoded using v3 encoding. You may choose to make that explicit since you know we're putting a collection IN a collection. Some other drivers do not handle it gracefully.

src/collection.cpp
mpenick: That's a good point. The driver will negotiate with only a single protocol version so mixing is not possible.  The driver will refuse to connect to a lower protocol version node if a higher protocol version has been previously established (https://github.com/datastax/cpp-driver/blob/master/src/connection.cpp#L76).

include/cassandra.h
mpenick: That's a good idea.

src/batch_request.cpp
mpenick: yes, that's part of it. I don't know what's going to need protocol version in the future and passing the protocol version alone doesn't seem like a good option. I was thinking this could be use to pass  around a custom allocator when that's implemented.

I'm open to better ways of handling this and sugggests. I just cant' think of a better way of handle this right now.

mpenick: @aholmberg ^ What do you think?

aholmberg: That sounds like a reasonable approach. Just some other thoughts for brainstorming:

The slow path could become more common as the cpp driver catches up to bleeding edge...

Could there be a global default setting controlling the max protocol for default encoding (removing the penalty for clients that configure it to their version).

Another option would be to defer encoding until the message is going to be written. May be less attractive since this might require copying values in, though.

mikefero:
Not sure if '3' was accidentally omitted or not https://github.com/datastax/cpp-driver/pull/170/files#diff-9533c4e18e4de61640f9b5534d635c51R36.

src/batch_request.cpp
mpenick: @aholmberg That sounds like a good idea. I like the idea of adding a setting to collection for the default protocol encoding. It will always be the latest but can be downgraded.

RE: Deferring the encoding. This is almost (or just) as expensive as re-encoding because the amount of work to decode to re-encode is not that expensive. Only the size of each element needs to be extracted and that doesn't require a full decode of that value i.e. decode_int32() doesn't need to be used. Deferred encoding was the way it was implemented before so I think the slow path will be even noticed.

PR #171, opened 2015-06-01
CPP-265 - Added support for protocol v3 schema change event

mpenick:


aholmberg:
I see schema meta model for UDT, but it is not exposed in the API here.

PR #179, opened 2015-06-23
Support building with Clang 3.6

gahr:
Clang 3.6 warns about unused local typedefs and causes a compilation error to occur because of -Wall.
See build log at
http://beefy4.nyi.freebsd.org/data/head-amd64-default/p390348_s284712/logs/cassandra-cpp-driver-2.0.1.log


mpenick:
@gahr Thanks for the patch. What platform are you running clang on? I'm building with clang 3.6 on OS X and "-Wno-unused-local-typedef" is not a valid flag...

gahr:
It's actually 3.6.1

mpenick:
Thanks again. It was a bit more complicated than checking for the compiler version because Apple's clang version is at 6+. This is fixed in CMake 3.0 (http://www.cmake.org/cmake/help/v3.0/policy/CMP0025.html#policy:CMP0025), but we need to support older platforms. So I need to added detection for Apple's clang vs LLVM's clang: https://github.com/datastax/cpp-driver/commit/4be751c52181d8bffee5a9bfc9fc6359f343f803

gahr:
Thanks a lot!

PR #185, opened 2015-07-14
266 - Added client-side timestamps

mpenick:


mpenick:
This builds on this PR: https://github.com/datastax/cpp-driver/pull/186

PR #186, opened 2015-07-14
278 - Retry policies

mpenick:


include/cassandra.h
```diff
@@ -1155,6 +1203,24 @@ cass_cluster_set_tcp_keepalive(CassCluster* cluster,
                                cass_bool_t enabled,
                                unsigned delay_secs);
 
+/**
+ * Sets the retry policy used for all requests unless overridden by setting
+ * a retry policy on a statement or a batch.
+ *
+ * <b>Default:</b> default retry policy.
```
aholmberg: "default retry policy": is this explained somewhere?

include/cassandra.h
```diff
@@ -4720,6 +4829,122 @@ cass_result_has_more_pages(const CassResult* result);
...
+ *
+ * @public @memberof CassErrorResult
+ *
+ * @param[in] error_result
+ * @return The actual received responses for a read timeout, actual recieved
+ * acknowlegements for a write timeout or actual alive nodes for a unavailable
+ * error. Undefined for other error result types.
+ */
+CASS_EXPORT cass_int32_t
+cass_error_result_actual(const CassErrorResult* error_result);
```
aholmberg: ``cass_error_result_received_responses``?

include/cassandra.h
```diff
@@ -4720,6 +4829,122 @@ cass_result_has_more_pages(const CassResult* result);
...
+ *
+ * @public @memberof CassErrorResult
+ *
+ * @param[in] error_result
+ * @return The required responses for a read time, required acknowlegements
+ * for a write timeout or required alive nodes for an unavailable error result.
+ * Undefined for other error result types.
+ */
+CASS_EXPORT cass_int32_t
+cass_error_result_required(const CassErrorResult* error_result);
```
aholmberg: same as above

include/cassandra.h
```diff
@@ -5488,6 +5713,143 @@ cass_uuid_from_string_n(const char* str,
...
+ * This policy retries queries in the following cases:
+ * <ul>
+ *   <li>On a read timeout, if enough replicas replied but data was not received.</li>
+ *   <li>On a write timeout, if a timeout occurs while writing the distributed batch log</li>
+ *   <li>On unavailble, it will move to the next host</li>
+ * </ul>
+ *
+ * In all other cases the error will be returned.
+ *
+ * This policy always uses the queries original consistency level.
```
aholmberg: "uses the query's..."

include/cassandra.h
```diff
@@ -5488,6 +5713,143 @@ cass_uuid_from_string_n(const char* str,
...
+ * @public @memberof CassRetryPolicy
+ *
+ * @param[in] child_retry_policy
+ * @return Returns a retry policy that must be freed. NULL is returned if
+ * the child_policy is a logging retry policy.
+ *
+ * @see cass_retry_policy_free()
+ */
+CASS_EXPORT CassRetryPolicy*
+cass_retry_policy_logging_new(CassRetryPolicy* child_retry_policy);
```
aholmberg: Good idea

include/cassandra.h
mpenick: The problem is it different for every error type. It's not always "received responses" it could be "number of alive nodes" or "number of acknowledgements". I was trying to avoid adding a new function for every type because it's really the same-ish value.

PR #188, opened 2015-07-16
152 - Idle connection heartbeat

mpenick:


mpenick:
Don't merge this yet because this is broken on older compilers. I'm thinking about moving the pending reconnect stuff into pool.

include/cassandra.h
```diff
@@ -1155,6 +1155,37 @@ cass_cluster_set_tcp_keepalive(CassCluster* cluster,
...
+ *
+ * @param[in] cluster
+ * @param[in] interval_ms Use 0 to disable heartbeat messages
+ */
+CASS_EXPORT void
+cass_cluster_set_connection_heartbeat_interval(CassCluster* cluster,
+                                               unsigned interval_secs);
+
+/**
+ * TODO
```
aholmberg: description

include/cassandra.h
```diff
@@ -1155,6 +1155,37 @@ cass_cluster_set_tcp_keepalive(CassCluster* cluster,
...
+ *
+ * <b>Default:</b> 60 seconds
+ *
+ * @public @memberof CassCluster
+ *
+ * @param[in] cluster
+ * @param[in] interval_ms
+ */
+CASS_EXPORT void
+cass_cluster_set_connection_idle_interval(CassCluster* cluster,
```
aholmberg: curious why there needs to be two settings here. I think of heartbeat interval as "send a request if the connection is idle for x seconds".

src/connection.cpp
```diff
@@ -750,13 +856,47 @@ void Connection::send_initial_auth_response() {
...
+    heartbeat_timer_.start(loop_,
+                           1000 * config_.connection_heartbeat_interval_secs(),
+                           this, on_heartbeat);
   }
 }
 
+void Connection::on_heartbeat(Timer* timer) {
+  Connection* connection = static_cast<Connection*>(timer->data());
+
+  if (connection->idle_time_ms_ == 0) {
```
aholmberg: I originally read this as "amount of idle time". Would it be appropriate to call it ``idle_start_time_`` to help dummies like me understand that it is absolute time?

include/cassandra.h
mpenick: I think a better name is "idle_timeout". A connection is terminated no heartbeat responses are received within this timeout period. A description would probably help.

PR #190, opened 2015-07-17
254 - Expose raw paging state

mpenick:


include/cassandra.h
```diff
@@ -4718,6 +4734,23 @@ cass_result_first_row(const CassResult* result);
 CASS_EXPORT cass_bool_t
 cass_result_has_more_pages(const CassResult* result);
 
+/**
+ * Gets the raw paging state from the result. The paging state is bound to the
+ * lifetime of the result object. If paging state needs to live beyond the
+ * lifetime of the result object it must be copied.
```
aholmberg: Do we need to mention the standard warning that paging state could be spoofed, and should not be exposed to untrusted clients?

mpenick: Good idea.

PR #193, opened 2015-07-27
Add timestamp conversion in cass_uuid_min/max_from_time() function to…

kirhist:
… be consistent with cass_uuid_gen_from_time() function. CPP-283.
In current code cass_uuid_min/max_from_time() do not convert timestamp to internal format which is inconsistent with the behavior of cass_uuid_gen_from_time() that uses from_unix_timestamp(() function for conversion. This commit adds same conversion to cass_uuid_min/max_from_time()  functions.


mpenick:
Thanks!

PR #194, opened 2015-07-29
266-tests: Added client-side timestamp tests

mpenick:


mikefero:
Added C* version checking so when integration tests are run using < C* 2.1, the tests will not be marked as failure.

Should we add the edge cases where the CQL timestamp should override the cass_cluster_set_timestamp_gen and/or cass_statement_set_timestamp settings?

PR #195, opened 2015-07-31
CPP-253 Ensure entire pool is not removed with partial connections

mikefero:
Removed old code and comment from outage test (fixed in CPP-140)

### DIFF
--- a/test/integration_tests/src/test_pool.cpp
+++ b/test/integration_tests/src/test_pool.cpp
@@ -22,12 +22,34 @@
 #include "test_utils.hpp"
 #include "cluster.hpp"
 
+#include <uv.h>
+
 #include <boost/scoped_ptr.hpp>
+#include <boost/thread.hpp>
 #include <boost/test/unit_test.hpp>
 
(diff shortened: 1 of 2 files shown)


mikefero:
Added partial reconnect into the test and squashed

test/integration_tests/src/test_pool.cpp
```diff
@@ -125,4 +147,93 @@ BOOST_AUTO_TEST_CASE(connection_spawn)
...
+ * @since 2.1.0
+ * @test_category connection:connection_pool
+ * @jira_ticket CPP-253 [https://datastax-oss.atlassian.net/browse/CPP-253]
+ */
+BOOST_AUTO_TEST_CASE(dont_recycle_pool_on_timeout) {
+  // Add a second node
+  ccm->bootstrap(2);
+
+  // Create the connection interruption data
+  ConnectionInterruptionData* ci_data = new ConnectionInterruptionData();
```
mpenick: `ci_data` can live on the stack.

PR #199, opened 2015-08-03
CPP-152: Integration test for handling idle connection failures (heartbeat)

mikefero:


test/integration_tests/src/test_heartbeat.cpp
```diff
@@ -0,0 +1,130 @@
...
+ * cluster
+ *
+ * @since 2.1.0
+ * @jira_ticket CPP-152
+ * @test_category connection:heartbeat
+ */
+BOOST_AUTO_TEST_CASE(interval) {
+  // Heartbeat disabled
+  cass_cluster_set_connection_heartbeat_interval(cluster, 0);
+  test_utils::CassLog::reset("Heartbeat request timed out on host " + conf.ip_prefix());
```
mpenick: Should this be `"Heartbeat completed on host " + conf.ip_prefix()` ?

mikefero: Good catch ... made correction.

PR #207, opened 2015-08-20
Adapt Debian packaking to jessie-backports package names.

thuovila:
To package the cpp-driver on Debian Jessie, the libuv driver from jessie-backports is required. There the library is packaged with the name "libuv1". These two changes adapt the Debian packaking to take the name into consideration.


thuovila:
Needs more work

PR #209, opened 2015-08-20
271 - Fixed infinite loop in HDR histogram

mpenick:


aholmberg:
Any plans to contribute back to the original project?

PR #214, opened 2015-08-25
291/292 - Support for custom payloads and warnings

mpenick:


include/cassandra.h
```diff
@@ -6052,6 +6184,70 @@ cass_retry_policy_free(CassRetryPolicy* policy);
...
+ *
+ * @public @memberof CassCustomPayload
+ *
+ * @param[in] item_count
+ * @return Returns a custom payload that must be freed.
+ *
+ * @see cass_custom_payload_free()
+ */
+CASS_EXPORT CassCustomPayload*
+cass_custom_payload_new(size_t item_count);
```
aholmberg: Why specify item count if we're going to use append semantics?

src/request.hpp
```diff
@@ -25,12 +25,31 @@
...
 
 namespace cass {
 
 class Handler;
 class RequestMessage;
 
+class CustomPayload : public RefCounted<CustomPayload> {
+public:
+  CustomPayload(size_t item_count)
+    : items_(item_count) { }
```
aholmberg: If we construct a vector with ``item_count`` then append, I think we're storing ``item_count`` empty buffers ahead of the actual items.

src/request.cpp
```diff
@@ -0,0 +1,75 @@
...
+  Buffer buf(sizeof(uint16_t) + name_length + sizeof(int32_t) + value_size);
+  size_t pos = buf.encode_string(0, name, name_length);
+  buf.encode_bytes(pos, reinterpret_cast<const char*>(value), value_size);
+  items_.push_back(buf);
+}
+
+int32_t CustomPayload::encode(BufferVec* bufs) const {
+  int32_t length = sizeof(uint16_t);
+  Buffer buf(sizeof(uint16_t));
+  assert(items_.size() % 2 == 0);
```
aholmberg: Why are we requiring an even number of items?

src/request.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+void cass_custom_payload_free(CassCustomPayload* payload) {
+  payload->dec_ref();
+}
+
+} // extern "C"
+
+namespace cass {
+
+void CustomPayload::append(const char* name, size_t name_length, const uint8_t* value, size_t value_size) {
```
aholmberg: I know the protocol encoding supports this, but 'append' semantics give me pause. Is there a reason we don't handle it as a mapping and use `set(key, value)` semantics instead? Wondering if this will be confusing with the payload handled as a map on the server side.

src/request.cpp
```diff
@@ -0,0 +1,75 @@
...
+  size_t pos = buf.encode_string(0, name, name_length);
+  buf.encode_bytes(pos, reinterpret_cast<const char*>(value), value_size);
+  items_.push_back(buf);
+}
+
+int32_t CustomPayload::encode(BufferVec* bufs) const {
+  int32_t length = sizeof(uint16_t);
+  Buffer buf(sizeof(uint16_t));
+  assert(items_.size() % 2 == 0);
+  buf.encode_uint16(0, items_.size() / 2);
```
aholmberg: I think there is opportunity for silent overflow here. Do we want to verify size and error out during ``cass_custom_payload_append``?

PR #220, opened 2015-09-14
304 - Adding VS 2015 build support and updating library dependency versions

mikefero:


test/integration_tests/src/test_control_connection.cpp
```diff
@@ -51,7 +51,7 @@ struct ControlConnectionTests {
       } else {
         CassString message;
         cass_future_error_message(future.get(), &message.data, &message.length);
-        BOOST_MESSAGE("Failed to query host: " << std::string(message.data, message.length));
+		BOOST_TEST_MESSAGE("Failed to query host: " << std::string(message.data, message.length));
```
mpenick: Tabs?

PR #228, opened 2015-09-21
Detect Apple's clang for warning flags

richardkchapman:
Older versions of Apple's of clang report the upstream clang version in
parentheses after the apple version - newer versions do not report the
upstream clang version at all.

This fix is required to be able to build with latest XCode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -446,16 +446,18 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
 
   # Detect if this is Apple's clang
   execute_process(COMMAND ${CMAKE_CXX_COMPILER} -v ERROR_VARIABLE CASS_CLANG_VERSION_INFO)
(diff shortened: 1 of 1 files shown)


mpenick:
Thanks a ton for finding this issue. What do you think about using "-Wno-unknown-warning-option"?

https://github.com/datastax/cpp-driver/commit/d68233f1c7a54094b114f8add6393b6c0c4da5da

Seems to work on Apple clang 6.1.0 (Xcode 6) and older versions of clang on Linux. I'll be testing Xcode 7 shortly (installing).

mpenick:
Verified that it also work with Apple clang 7.0.0.

richardkchapman:
-Wno-unknown-warning-option looks like it is a more general fix for the issue - so long as there are no compilers out there so old that they don't recognise that option, of course.

mpenick:
Landed in https://github.com/datastax/cpp-driver/commit/e6a4f9505b032bc26d44de0291eba21a1edc48f5. Going with `-Wno-unknown-warning-option ` for now.

PR #230, opened 2015-10-02
Replace LZCNT with BSR instruction in hdr_histogram

SeverinLeonhardt:
The LZCNT instruction produced by __lzcnt on Windows is not available
on many CPUs in use today. BSR is a safe replacement requiring only
minimal additional work.

### DIFF
--- a/src/third_party/hdr_histogram/hdr_histogram.cpp
+++ b/src/third_party/hdr_histogram/hdr_histogram.cpp
@@ -23,21 +23,36 @@
 
 #include "hdr_histogram.hpp"
 
-inline int32_t hdr_clz64(uint64_t x) {
+/**
+ * Get the most significant bit set.
+ *
+ * The result is a one-based index, zero means no bits are set.
+ */
+inline int32_t hdr_bsr64(uint64_t x) {
 #if defined(_MSC_VER)
+  unsigned long index;
 #  if defined(_M_AMD64)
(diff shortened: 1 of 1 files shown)


mpenick:
@SeverinLeonhardt  Thanks for researching and getting a pull request together. I just had one issue compiling on Windows, otherwise, LGTM.

src/third_party/hdr_histogram/hdr_histogram.cpp
```diff
@@ -23,21 +23,36 @@
...
+  char isNonzero = _BitScanReverse64(&index, x);
+  if (isNonzero)
+    return index + 1;
+  else
+    return 0;
 #  else
   // On 32-bit this needs to be split into two operations
-  int32_t lz = (int32_t)__lzcnt((unsigned long)(x >> 32));
-  if (lz == 32) {
+  const char isNonzero = _BitScanReverse(&index, (unsigned long)(x >> 32));
```
mpenick: Should this be "const"?

mpenick:
I noticed another issue too that's unrelated to the changes you made. For GCC-based compilers `__builtin_clzll(0)` is undefined so the code needs to check that value `x` for `0` (and `return 0`).

SeverinLeonhardt:
Updated the pull request as you've suggested.

mpenick:
Thanks for updating. Looks good.

PR #232, opened 2015-10-09
148 - Whitelist load balancing policy

mikefero:


src/whitelist_policy.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include "load_balancing.hpp"
+#include "host.hpp"
+#include "scoped_ptr.hpp"
+
+namespace cass {
+
+class WhitelistPolicy : public ChainedLoadBalancingPolicy {
+public:
+  typedef std::list<std::string> ContactPointList;
```
mpenick: I'd make this a `std::vector<>` or a `std::set<>`. Ideally, this would be a hash table set, but we'd have to either build one or use sparse hash. `std::list<>` is only really good for adding/removing from a large list in the middle of the list.

src/cluster.cpp
```diff
@@ -288,6 +287,37 @@ void cass_cluster_set_latency_aware_routing_settings(CassCluster* cluster,
...
+                                             hosts_length);
+}
+
+void cass_cluster_set_whitelist_routing_hosts_n(CassCluster* cluster,
+                                                const char* hosts,
+                                                size_t hosts_length) {
+  if (hosts_length == 0) {
+    cluster->config().whitelist_hosts().clear();
+  } else {
+    std::istringstream stream(std::string(hosts, hosts_length));
```
mpenick: `cass_cluster_set_contact_list()` uses this exact same functionality. It might make sense to extract this into its own function private (file static) function.

PR #233, opened 2015-10-14
301 - Schema metadata API refactor

mpenick:


include/cassandra.h
```diff
@@ -1515,203 +1509,428 @@ cass_session_get_schema(CassSession* session);
...
+/**
+ * Gets the id of the table.
+ *
+ * @public @memberof CassTableMeta
+ *
+ * @param[in] table_meta
+ * @return The UUID id of the table.
+ */
+CASS_EXPORT CassUuid
+cass_table_meta_id(const CassTableMeta* table_meta);
```
mikefero: For C* v1.2.x an 2.0.x `cf_id` does not exist resulting in the return of a uninitialized id_ (https://github.com/datastax/cpp-driver/pull/233/files#diff-f30001b0d6d7abd11f9acc768a8b0737R240); the CassUuid struct does not define a constructor to to initialize either. Should we define a CASS_UUID_INVALID in the API header and initialize CassUuid as the invalid define?

Or we could document that the meta id is only available for C* 2.1+.

aholmberg:
I'm seeing some kind of corruption in the output of the example:

```
Error: Cannot add existing keyspace "examples"
Keyspace "annot ":
....
Table "schema_
              ":
```
The names of each are messed up. Just did some very basic debugging and they seem to be coming back correctly. Have not dug in further... posting here for corroboration.

mpenick:

PR #234, opened 2015-10-16
Address::compare() fixed

ybainier:
Some IPv4 addresses could not be found in `IOWorker::PoolMap` instance because  `Address::compare()` method was broken on *Windows x64*.

### DIFF
--- a/src/address.cpp
+++ b/src/address.cpp
@@ -139,12 +139,12 @@ std::string Address::to_string(bool with_port) const {
   return ss.str();
 }
 
-int Address::compare(const Address& a) const {
+std::int64_t Address::compare(const Address& a) const {
   if (family() != a.family()) {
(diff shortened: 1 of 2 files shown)


mpenick:
@ybainier Great find! That's a huge unsigned subtract fail. Thanks.

I would almost prefer using: `if  (a != b) return a < b ? -1 : 1` then have a default `return 0` at the end. That way we don't have to use a 64-bit type (the driver is being used on 32-bit platforms too) What do you think?

ybainier:
That's clever and looks better.

mpenick:
Change landed in https://github.com/datastax/cpp-driver/commit/b2960302bc8182497e01a5b99127a15dfb4431b6. Thanks a ton for finding the issue and providing a fix.

PR #237, opened 2015-10-16
308 - Change default consistency from ONE to QUORUM

mpenick:


src/batch_request.hpp
```diff
@@ -38,7 +38,7 @@ class BatchRequest : public RoutableRequest {
   BatchRequest(uint8_t type_)
       : RoutableRequest(CQL_OPCODE_BATCH)
       , type_(type_)
-      , consistency_(CASS_CONSISTENCY_ONE) {}
+      , consistency_(CASS_CONSISTENCY_QUORUM) { }
```
aholmberg: I think this should be LOCAL_QUORUM.

src/request.hpp
```diff
@@ -59,7 +59,7 @@ class Request : public RefCounted<Request> {
 
   Request(uint8_t opcode)
       : opcode_(opcode)
-      , consistency_(CASS_CONSISTENCY_ONE)
+      , consistency_(CASS_CONSISTENCY_QUORUM)
```
aholmberg: Same

aholmberg:
Should this be updated? https://github.com/datastax/cpp-driver/blob/308/src/dc_aware_policy.cpp#L62

aholmberg:
doc update: https://github.com/datastax/cpp-driver/blob/master/include/cassandra.h#L2155

PR #240, opened 2015-10-20
Fix documentation of default values

SeverinLeonhardt:
Some documented values in cassandra.h didn't match the actual values
in config.hpp.

The pending requests high water mark referred to max_connections_per_host
suggesting it is calculated from that value. This is not the case, it will
always be 128 * 2.


mpenick:
Thanks for updating.

PR #241, opened 2015-10-20
318 - Removed upper bound protocol validation

mpenick:
It will automatically be negotiated to a valid value using the control
connection and it's one more place in the code that needs to be updated.


aholmberg:
+1

PR #242, opened 2015-10-21
320 - Fixed IPv4 address comparison

mpenick:
Special thanks to https://github.com/ybainier for finding and providing
the initial fix for the issue.

The original PR: https://github.com/datastax/cpp-driver/pull/234


aholmberg:
+1

PR #244, opened 2015-10-26
Added field by name functions to funciton/aggregate meta and updated docs

mpenick:


mikefero:
+1. Will merge into 295-test branch after merge with master to update integration test

PR #249, opened 2015-11-17
329 - Atomic [MSVC Intrinsics] fails to compile; 32-bit only

mikefero:
Also addressed threshold issues for 32-bit slowness of metrics

### DIFF
--- a/src/atomic/atomic_intrinsics_msvc.hpp
+++ b/src/atomic/atomic_intrinsics_msvc.hpp
@@ -55,32 +55,7 @@
 // 32-bit Windows compilation
 #ifndef _M_X64
 # pragma intrinsic(_InterlockedCompareExchange64)
-
-__inline __int64 _InterlockedExchangeAdd64(__int64 volatile *storage,
-                                           __int64 value) {
-  // Short circuit
-  if (value == 0) {
-    return *storage;
-  }
-
-  // Perform operation
-  __int64 old_value;
-  do {
-    old_value = *storage;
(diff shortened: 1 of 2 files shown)


src/atomic/atomic_intrinsics_msvc.hpp
```diff
@@ -52,6 +52,37 @@
...
+  // Short circuit
+  if (value == 0) {
+    return *storage;
+  }
+
+  // Perform operation
+  __int64 old_value;
+  do {
+    old_value = *storage;
+  } while(_InterlockedCompareExchange64(storage, (*storage + value), *storage) != old_value);
```
mpenick: This should be: `_InterlockedCompareExchange64(storage, (old_value + value), old_value) != old_value`, otherwise, you could get a different value for each `*storage` as it's changed by another thread.

src/atomic/atomic_intrinsics_msvc.hpp
```diff
@@ -52,6 +52,37 @@
...
+  return old_value;
+}
+
+__inline __int64 _InterlockedExchange64(__int64 volatile *storage,
+                                        __int64 value) {
+  __int64 ret = *storage;
+  _InterlockedCompareExchange64(storage, value, *storage);
+  return ret;
+}
```
mpenick: I think we need to do something like this because we don't want to give up on the exchange if some other thread changed the value of storage while we attempting to update.
```c
__int64 ret;
do {
   ret = *storage;
} while(_InterlockedCompareExchange64(storage, value, ret) != ret); 
```

PR #250, opened 2015-11-18
'All hosts in current policy attempted and were either unavailable or failed' error

ureyni:
hi..
i write to c++ program using cpp-driver-master.
i get to error message this 'All hosts in current policy attempted and were either unavailable or failed'.
i dont found knowledge anywhere 
What could be the cause of the error.

--Cassandra is one node server.


ureyni:
i solved this problems.
KeySpace drop and create but 
one change cql command , 
replication_factor is 1 instead of 3 
CREATE KEYSPACE dovecot
  WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 3->1 };

thanks

PR #251, opened 2015-11-20
Fix compile error in CCMBridge on Windows

SeverinLeonhardt:
In contrast to the other functions _read is declared in io.h.


mpenick:
Thanks!

PR #253, opened 2015-11-30
330 - Support for Cassandra 3.0 schema metadata

mpenick:


aholmberg:
Will ``examples/schema_metadata`` be updated with this, or another ticket? I get various errors when running against 2.2 and 3.0 (different depending on Cassandra version).

mpenick:
Good call. I didn't run the example. I was only running tests...

src/control_connection.cpp
```diff
@@ -683,11 +725,11 @@ void ControlConnection::on_refresh_keyspace(ControlConnection* control_connectio
 
 void ControlConnection::refresh_table(const StringRef& keyspace_name,
                                       const StringRef& table_name) {
-  std::string cf_query(SELECT_COLUMN_FAMILIES);
+  std::string cf_query(SELECT_COLUMN_FAMILIES_20);
```
aholmberg: Do we need a ``_30`` version of this refresh? Is that TBD?

aholmberg:
+1 looks good

PR #256, opened 2016-01-07
FreeBSD defines major/minor

gahr:


mpenick:
Thanks! This could be a problem on other *nix platforms (not just FreeBSD) and to make it more complex sometimes they're not macro, but inlined functions. I wonder if we should rename those params / methods.

gena01:
@mpenick Seems that Alpine Linux has a similar problem (need it for Docker image/container...)

mpenick:
Adding fix in the next release: https://datastax-oss.atlassian.net/browse/CPP-345

mpenick:
Fix: https://github.com/datastax/cpp-driver/pull/262 up for review.

mikefero:
Thanks for the contribution @gahr; #262 has been merged into master.

gahr:
Thanks @mikefero!

PR #259, opened 2016-02-23
344: Make libssh2 optional when building tests

mikefero:
Since CCM commands can now be run locally libssh2 can be considered optional. This change will present a status message if -DCASS_USE_LIBSSH2=ON and cannot be found; indicating libssh2/remote commands will be disabled. This change also allows for the building of either the integration (libssh2 optional), unit tests, or both.


CMakeLists.txt
```diff
@@ -60,6 +65,8 @@ option(CASS_BUILD_STATIC "Build static library" OFF)
 option(CASS_BUILD_EXAMPLES "Build examples" OFF)
 option(CASS_BUILD_DOCS "Build documentation" OFF)
 option(CASS_BUILD_TESTS "Build tests" OFF)
+option(CASS_BUILD_INTEGRATION_TESTS "Build tests" OFF)
+option(CASS_BUILD_UNIT_TESTS "Build tests" OFF)
```
mpenick: Maybe this should say "Build integration tests" and "Build unit tests"?

mikefero: Good call; that is why you shouldn't copy and paste.

PR #260, opened 2016-02-24
331 - Support for materialized view metadata

mpenick:


mpenick:
Good catch! That would have been bad.

aholmberg:
+1

PR #262, opened 2016-02-26
345 - Changed major/minor to major_version/minor_version to avoid naming co…

mpenick:
…nflicts on some platforms


mikefero:
+1; verified on FreeBSD 10.2

mikefero:
Similar changes need to be made in CCM bridge; will update in PR #259 due to potential conflicts.

PR #264, opened 2016-02-29
281 - Handle error when connecting with keyspace that doesn't exist

mpenick:
Also, 690c7d0 fixes a bottleneck I found when benchmarking.


aholmberg:
+1 changes

One thing I noticed while experimenting: examples/callback now hangs if keyspace 'examples' does not exist  because it was doing ``cass_session_connect_keyspace`` without knowing the keyspace exists, and the exit handling was incorrect.

I posted #267 for this branch if you want it.

PR #265, opened 2016-02-29
296 - Docs: Started adding Cassandra version to API features

mpenick:
This will be useful for denoting what features are supported with what version of Cassandra.


mikefero:
+1

PR #267, opened 2016-03-02
Fix callback example

aholmberg:
- Don't connect_keyspace (in case it's not there)
- Fix exit handling for failed session connect


mpenick:
+1 Thanks. Merge at will.

PR #271, opened 2016-03-08
Fix: Added missing materialized view function and fix clustering key...

mpenick:
…nd fix cluster key order positions


mikefero:
+1

PR #273, opened 2016-03-15
Added performance tips to docs

mpenick:


topics/configuration/README.md
```diff
@@ -146,5 +146,71 @@ cass_cluster_set_connection_idle_timeout(cluster, 120);
...
+```c
+cass_cluster_set_write_bytes_low_water_mark(cluster, 512 * 1024);
+cass_cluster_set_write_bytes_high_water_mark(cluster, 1024 * 1024);
+```
+
+#### Use [paging] when retrieving large result sets
+
+Using a large page size or a very high `LIMIT` clause can cause your application
+to delay for each individual request. The driver's paging mechanism can be use
+to decrease the latency of individual requests.
```
stamhankar999: use -> used

topics/configuration/README.md
```diff
@@ -146,5 +146,71 @@ cass_cluster_set_connection_idle_timeout(cluster, 120);
...
+
+#### Choose a lower consistency level
+
+Ultimately, choosing a consistency level is a trade-off between consistency and
+availability. Performance should not be a large deciding factor when choosing a
+consistency level. However, it can affect high-percentile latency numbers
+because requests with consistency levels greater than `ONE` can cause requests
+wait for one or more nodes to response back to the coordinator node before a
```
stamhankar999: wait => to wait

topics/configuration/README.md
```diff
@@ -146,5 +146,71 @@ cass_cluster_set_connection_idle_timeout(cluster, 120);
...
+request can complete. In multi-datacenter configures consistency levels such as
```
stamhankar999: configures => configuration,

topics/configuration/README.md
```diff
@@ -146,5 +146,71 @@ cass_cluster_set_connection_idle_timeout(cluster, 120);
...
+`EACH_QUORUM` can cause a request to wait for replication across a slower cross
+datacenter network link.  More information about setting the consistency level
+can be found [here](http://datastax.github.io/cpp-driver/topics/basics/consistency/).
+
 [`allow_remote_dcs_for_local_cl`]: http://datastax.github.io/cpp-driver/api/CassCluster/#1a46b9816129aaa5ab61a1363489dccfd0
 [`OPTIONS`]: https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v3.spec#L278-L282
```
stamhankar999: Ok, I realize this was already here, but it's not entirely safe to refer to files on trunk, since over time they may change (e.g. what if some day the v3 protocol is dropped, and this file is removed from trunk?). Refer to a tag instead, so the link is guaranteed to be valid.

stamhankar999:
Most of this seems like a good candidate for putting in a "non-driver-specific useful info" section/doc of docs.datastax.com. Might want to consider asking Jim or Paul if they can create such a section/doc. Then you (really all of the drivers) can have their README's reference said doc.

mpenick:
Thanks for the feedback. Great tip on not referring to master/trunk.

PR #276, opened 2016-03-23
Handle permissive errors for non GNU compilers

mikefero:


jeremyong:
+1

gokhandincer:
Thank you. it is work on alpine

PR #282, opened 2016-05-02
Resolve multiple addresses for each domain name.

mkm85:
Resolve multiple addresses for each domain name. Before only one IP was returned even if the domain name resolves to several IP addresses. This is useful if a domain name resolves to all the the client seednodes.

The code is tested but no test is included since it would rely on a nameserver which cannot be _simply_ and reliably provided to an integration test.

### DIFF
--- a/src/resolver.hpp
+++ b/src/resolver.hpp
@@ -49,7 +49,7 @@ public:
   bool is_success() { return status_ == SUCCESS; }
   bool is_timed_out() { return status_ == FAILED_TIMED_OUT; }
   Status status() { return status_; }
-  const std::vector<Address>& addresses() const { return addresses_; }
+  const AddressVec& addresses() const { return addresses_; }
   T& data() { return data_; }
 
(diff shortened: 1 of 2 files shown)


mpenick:
Thanks for the pull request. I've create a JIRA issue to track it: https://datastax-oss.atlassian.net/browse/CPP-364. I'll pull this into master right after 2.4 releases.

src/resolver.hpp
```diff
@@ -49,7 +49,7 @@ class Resolver {
   bool is_success() { return status_ == SUCCESS; }
   bool is_timed_out() { return status_ == FAILED_TIMED_OUT; }
   Status status() { return status_; }
-  const Address& address() const { return address_; }
+  const std::vector<Address>& addresses() const { return addresses_; }
```
mpenick: There's an `AddressVec` typedef in `address.h`. I'd prefer this use that.

PR #284, opened 2016-05-06
Added docs to clarify the immutability of the session's configuration

mpenick:


AlexisWilke:
JIRA references in link with this entry:
https://datastax-oss.atlassian.net/browse/CPP-362
https://datastax-oss.atlassian.net/browse/CPP-300

PR #288, opened 2016-05-16
367 - Fix: Session connect hangs because of "UP" event

mpenick:
Adding a connection pool as a result of an event before the driver is
done with its connection process causes the pending pool count to never
decrement to zero. Events are no longer able to add or remove pools
until the control connection ready.


mikefero:
👍

PR #290, opened 2016-05-23
Fix: Adding valid data type to null values caused segfault in metadata

mpenick:
- Adding a data type to "null" values (commit:
  bf2d53f3542fcdbb41d9cff113b8c56d9931cd10) changed an assumption about
  how a "null" (or valid) values could be determined.


mikefero:
👍

PR #297, opened 2016-06-10
300: Added per-statement and per-batch request timeouts

mpenick:


src/handler.cpp
```diff
@@ -140,6 +140,15 @@ void Handler::set_state(Handler::State next_state) {
       assert(false && "Invalid request state");
       break;
   }
+
+}
+
+uint64_t Handler::request_timeout_ms(const Config& config) const {
+  uint64_t request_timeout_ms = request_->request_timeout_ms();
+  if (request_timeout_ms == CASS_UINT64_MAX) {
```
mikefero: 0 might be better than `CASS_UINT64_MAX` to set assign the default cluster configuration timeout.

**`uv_start_timer`** indicates:

> If timeout is zero, the callback fires on the next event loop iteration. If repeat is non-zero, the callback fires first after timeout milliseconds and then repeatedly after repeat milliseconds.

Reason being is a user might want to set the max timeout in order to "wait forever" and might by default set that value to the max.

mikefero:
Besides the one comment; 👍

PR #302, opened 2016-06-28
381- Fix: Hang in schema change handler cause by server error

mpenick:


src/schema_change_handler.cpp
```diff
@@ -109,9 +109,8 @@ void SchemaChangeHandler::on_set(const ResponseMap& responses) {
       has_error = true;
     }
   }
-  if (has_error) return;
 
-  if (has_schema_agreement(responses)) {
+  if (!has_error && has_schema_agreement(responses)) {
```
aholmberg: Not totally necessary, but it might be useful to log something more than `DEBUG`/schema not up-to-date if `has_error` (below, line 126).

aholmberg:
+1 with minor, optional comment

src/schema_change_handler.cpp
mpenick: @aholmberg An error is printed for each of the error responses in https://github.com/datastax/cpp-driver/blob/381/src/error_response.cpp#L210-Lundefined

PR #303, opened 2016-07-01
Make sure our in-source include directories come first.

gahr:
This avoids including a header from a previously installed version of
cpp-driver.


mpenick:
Thanks for the commit. Very helpful.

PR #305, opened 2016-07-11
token aware routing broken + code complexity of map_replicas too high with vnodes

ecourreges-orange:
- token aware routing is broken
- complexity of map_replicas is too high (proportionnal to number of tokens) and leads to timeout on reconnect of control connection, if timeout is set at <3s for a cluster of >10 nodes with vnodes activated.

### DIFF
--- a/src/control_connection.cpp
+++ b/src/control_connection.cpp
@@ -519,6 +519,7 @@ void ControlConnection::on_query_meta_schema(ControlConnection* control_connecti
 
   Session* session = control_connection->session_;
 
+  // this clears token map and breaks token aware routing
   session->metadata().clear_and_update_back();
 
   bool is_initial_connection = (control_connection->state_ == CONTROL_STATE_NEW);
@@ -616,6 +617,7 @@ void ControlConnection::refresh_node_info(SharedRefPtr<Host> host,
   }
 }
 
+  // duplicate code with on_refresh_node_info_all
 void ControlConnection::on_refresh_node_info(ControlConnection* control_connection,
                                              const RefreshNodeData& data,
                                              Response* response) {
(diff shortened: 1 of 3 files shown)


mpenick:
@ecourreges-orange Thanks for your feedback. This is something I'll starting digging into first thing tomorrow. First, I'll get token-awareness  working properly then I'll look into reducing the complexity of `map_replicas()`.

ecourreges-orange:
Sorry, I don't know where to post issues/feature requests, but here are the ones related to this Pull Request:
-	The driver should connect randomly to the contact_points at startup, not in lexicographical order => this will limit the impact of having the control_connection host/node down. The java driver already does that.
-	The useSchema and tokenAware functionalities should be independent, as useSchema induces unneeded cost when just using tokenAware. The driver should only get the keyspace list.
-	In case of connection_control reconnect, map_replicas() should be called only once at the end, not for every node
-	Complexity N of tokens_to_replicas should be reduced and tested on a cluster of at least 10 nodes with vnodes (i.e. >=2560 token ranges).
-	Request Timeout should be decorrelated from control_connection actions => in the case above, if you set requestTimeout too low, you end up with an infinite loop (with 100% CPU) of control_connection reconnects that do map_replicas()+timeout.
-	on_query_meta_schema() should not clear the token_map built in on_query_hosts()
-	All of this must be unit tested
Thank you.
Regards,
Emmanuel.

mpenick:
Thanks again. For future issues you can use JIRA: https://datastax-oss.atlassian.net/, but feel free to continue using GitHub. I've created this issue: https://datastax-oss.atlassian.net/browse/CPP-389 to track the problems with token awareness (and also linked related issues). 

The randomizing contact points improvement is tracked in this issue: https://datastax-oss.atlassian.net/browse/CPP-193

mpenick:
Issues addressed @ 4c744dc6537596de05ceb1c9da4ff8da67e9f586. Please let use know if this resolves your issues.

PR #312, opened 2016-08-03
Fix version check in schema_metadata/lookup test

SeverinLeonhardt:
The comment of the test states this test is meant for Cassandra 2.2+ yet the check makes it run against version 2.1 as well.


mpenick:
Good catch. Thanks.

PR #313, opened 2016-08-03
Minor fixes in CppDriver.cmake

SeverinLeonhardt:
As far as I can tell boost date_time, filesystem, log / log_setup and regex are not used anywhere.
I was able to build the driver after removing the components from find_package.


mikefero:
Thank you for the commit; always nice to have less dependencies.

Your changes have been merged along with documentation and Boost minimum version updates in 3ccb0f9

PR #316, opened 2016-08-25
Fix cass_uuid_from_string_n for arm

doughdemon:


mpenick:
Thanks for the patch. Fix with:  https://github.com/datastax/cpp-driver/pull/320/commits/f6c8e8e45046915377cabc56de3645ea9c1d5ee5

PR #320, opened 2016-09-06
Fix problems compiling on power8 due to char defaulting to unsigned

ghalliday:
Signed-off-by: Gavin Halliday <gavin.halliday@lexisnexis.com>

### DIFF
--- a/src/uuids.cpp
+++ b/src/uuids.cpp
@@ -121,7 +121,7 @@ CassError cass_uuid_from_string_n(const char* str,
   const char* pos = str;
   char buf[16];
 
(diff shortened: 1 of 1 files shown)


ghalliday:
On power8 (or any system that defaults char to unsigned) you get the following errors:

.../src/uuids.cpp:141:3: error: narrowing conversion of ‘-1’ from ‘int’ to ‘const char’ inside { } [-Werror=narrowing]

and 

.../src/uuids.cpp:151:30: error: comparison is always false due to limited range of data type [-Werror=type-limits]

mpenick:
Thanks! Good simple fix.

PR #321, opened 2016-09-08
403 - Fix: Random partitioner hash values should be in the range [1, 2^127-1]

mpenick:


mpenick:
Working on unit tests.

mpenick:
@aholmberg I actually fixed the issue and added unit tests.

aholmberg:
+1 here

PR #323, opened 2016-09-21
399 - Speculative execution

mpenick:


src/session.cpp
```diff
@@ -695,11 +701,16 @@ void Session::on_execute(uv_async_t* data) {
...
   bool is_closing = false;
 
-  RequestHandler* request_handler = NULL;
-  while (session->request_queue_->dequeue(request_handler)) {
-    if (request_handler != NULL) {
-      request_handler->set_query_plan(session->new_query_plan(request_handler->request(),
-                                                              request_handler->encoding_cache()));
+  RequestHandler* temp = NULL;
+  while (session->request_queue_->dequeue(temp)) {
+    RequestHandler::Ptr request_handler(temp);
```
stamhankar999: This could move down into the if and change the if condition to be `if (temp) {`.

src/cluster.cpp
```diff
@@ -421,6 +424,24 @@ CassError cass_cluster_set_use_randomized_contact_points(CassCluster* cluster,
...
+  }
+  cluster->config().set_speculative_execution_policy(
+        new cass::ConstantSpeculativeExecutionPolicy(constant_delay_ms,
+                                                     max_speculative_executions));
+  return CASS_OK;
+}
+
+CassError cass_cluster_set_no_speculative_execution_policy(CassCluster* cluster) {
+  cluster->config().set_speculative_execution_policy(
+        new cass::NoSpeculativeExecutionPolicy());
```
mpenick: In general, the driver avoids allocating memory during static initialization (Actually, we try to avoid any sort of complex static initialization) The main reason being that it prevents up from using a custom allocator to make those allocations when the default allocator is overridden.  We've had several users ask for this feature: https://datastax-oss.atlassian.net/browse/CPP-360

src/io_worker.cpp
```diff
@@ -135,38 +135,39 @@ void IOWorker::add_pool(const Host::ConstPtr& host, bool is_initial_connection)
...
-  Address address;
-  if (!request_handler->get_current_host_address(&address)) {
-    request_handler->on_error(CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
-                              "All hosts in current policy attempted "
-                              "and were either unavailable or failed");
-    return;
+bool IOWorker::execute(const RequestHandler::Ptr& request_handler) {
+  request_handler->inc_ref(); // Queue reference
+  if (!request_queue_.enqueue(request_handler.get())) {
+    request_handler->dec_ref();
```
mpenick: I do agree with the error proneness of manually inc/dec ref-counted object. I think we try to avoid this in most parts of the driver and I've attempted to make improvements in this release. If there are specific cases where you think it can be fixed let me know. There are a few cases where this might not be avoided:

1) Returning a reference to the external API. This is because the fact that it's a ref-counted type is lost.
2) In async code where the type is erased or it's added to a data structure that doesn't manage lifetime (e.g. an intrusively linked list). There's likely some improvement that could be made here in some places in `Connection`. 
3)  I've attempted to use `MCMPQueue<RequestHandler::Ptr>` instead of `MSMPQueue<RequestHandler*>`. The problem is the queue holds on to request memory until another request overwrites it or until the queue is destroyed. One solution is to proactively set the item to an empty object, `T()`, when dequeuing it, however, I am almost certain that would cause data races. The problem with keeping the memory alive is the size of requests are determine by the integrating application and could be quite large and there's an expectation that when a request is done that memory will be reclaimed.

The driver does not use exceptions  and limits the use of std library methods that throw exceptions.

src/io_worker.cpp
```diff
@@ -281,13 +282,17 @@ void IOWorker::on_execute(uv_async_t* async) {
 #endif
   IOWorker* io_worker = static_cast<IOWorker*>(async->data);
 
-  RequestHandler* request_handler = NULL;
+  RequestHandler* temp = NULL;
   size_t remaining = io_worker->config().max_requests_per_flush();
-  while (remaining != 0 && io_worker->request_queue_.dequeue(request_handler)) {
-    if (request_handler != NULL) {
+  while (remaining != 0 && io_worker->request_queue_.dequeue(temp)) {
+    RequestHandler::Ptr request_handler(temp);
```
mpenick: See above for why the queue uses `RequestHandler*` instead of `RequestHandler::Ptr`. 

The goal was to get the `RequestHandler` into a shared ptr as soon as possible to avoid error (in future changes). Also, the `SepeculativeExecution`constructor takes a shared ptr.

src/request_handler.cpp
```diff
@@ -16,24 +16,176 @@
...
+private:
+  SpeculativeExecution::Ptr speculative_execution_;
+};
+
+void PrepareCallback::on_internal_set(ResponseMessage* response) {
+  switch (response->opcode()) {
+    case CQL_OPCODE_RESULT: {
+      ResultResponse* result =
+          static_cast<ResultResponse*>(response->response_body().get());
+      if (result->kind() == CASS_RESULT_KIND_PREPARED) {
```
mpenick: The request has been chained and we're retrying the original request. We've attempted to run a prepared query and the server responded with `UNPREPARED` so we've wrapped the original request in a prepared request and are now returning to the original request.

src/session.cpp
mpenick: The goal was to get the RequestHandler into a shared ptr as soon as possible to avoid error (in future changes). Also, see above.

src/request_handler.cpp
```diff
@@ -169,70 +316,82 @@ void RequestHandler::on_result_response(ResponseMessage* response) {
...
 
+  switch(error->code()) {
     case CQL_ERROR_READ_TIMEOUT:
-      handle_retry_decision(response,
-                            retry_policy_->on_read_timeout(error->consistency(),
-                                                           error->received(),
-                                                           error->required(),
-                                                           error->data_present() > 0,
-                                                           num_retries_));
+      decision =  request_handler_->retry_policy()->on_read_timeout(request(),
```
mpenick: I'm pretty sure that Cassandra read requests are always idempotent.

src/cluster.cpp
mpenick: A potentially better optimization is to keep returning the same instance of `NoSpeculativeExecutionPlan` (vs policy) since it is allocated for ever request. I've attempted to do this (using custom `operator new` and `operator delete` without luck because multiple threads attempt to write the vtable simultaneously. That tiny allocation doesn't add very much overhead because is likely cached in the allocator anyway.

stamhankar999: Agreed, not a huge deal here. That said, you *could* create a NoSpeculativeExecutionPlan data member in NoSpeculativeExecutionPolicy's constructor and return that one plan object anytime a user asks for a plan from this policy object. So, you won't have one NoSpeculativeExecutionPlan object in the whole application, but rather O(# NoSpeculativeExecutionPolicy objects), which is still better than O(# active requests). In fact, most of the time you'll only have one cluster object, so # of NoSpeculativeExecutionPolicy is often 1.

src/request_handler.cpp
mpenick: "Read_timeout: Timeout exception during a read request."  

https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v3.spec#L982

It explicitly says "write request" and "read request". I'm pretty sure it doesn't have to do we a failed acknowledgement from a replica write as that would manifest as a `Write_timeout`.

src/cluster.cpp
mpenick: That can't be done because the calling code is responsible for deleting the plan and would incorrectly delete the shared instance.  We could use a ref-counted object for plans or figure out some custom allocation solution as I attempted before, but I think it's overkill.

stamhankar999: Agreed; not worth the trouble if it's even remotely non-trivial.

src/io_worker.cpp
stamhankar999: I think I'd like to have a longer discussion about this later; this is a broader question that isn't particularly related to speculative execution, so I'm cool with what you've got here.

src/io_worker.cpp
stamhankar999: Gotcha.

src/io_worker.cpp
```diff
@@ -281,13 +282,17 @@ void IOWorker::on_execute(uv_async_t* async) {
...
+    if (request_handler) {
+      request_handler->dec_ref(); // Queue reference
       io_worker->pending_request_count_++;
-      request_handler->set_io_worker(io_worker);
-      request_handler->retry();
+      request_handler->start_request(io_worker);
+      SpeculativeExecution::Ptr speculative_execution(new SpeculativeExecution(request_handler,
```
stamhankar999: Ahh, I see that now.

src/request_handler.cpp
stamhankar999: Ok.

src/request_handler.cpp
stamhankar999: Makes sense.

src/session.cpp
stamhankar999: 👍

stamhankar999:
Ok, I think you've addressed everything I wondered about. Merge at will! :)

PR #324, opened 2016-10-03
Ensure that the CMAKE_MODULE_PATH is set correctly even if this project is transitively linked

jeremyong:
In the case that the cpp-driver is built as a transitively linked library
(aka embedded in another CMake project), the module path will be augmented
relative to the project root instead of this project.


mpenick:
Thanks for the fix. Reviewing.

PR #329, opened 2016-12-15
Document how to run unit and integration tests

stamhankar999:
* Update docs.yaml to exclude an appveyor url from external link checking. The link is good, but documentor erroneously thinks the link is bad.

### DIFF
--- a/topics/testing/README.md
+++ b/topics/testing/README.md
@@ -1,6 +1,21 @@
 # Testing
 Before proceeding ensure the tests were built using the [build procedures].
 
+Both unit and integration tests use [Boost.Test](http://www.boost.org/doc/libs/1_62_0/libs/test/doc/html/index.html)
+and are packaged up into the `test/unit_tests/cassandra_unit_tests` and
+`test/integration_tests/cassandra_integration_tests` binaries, respectively. Run the test program with desired options:
+
+```
+# Run all integration tests
+test/integration_tests/cassandra_integration_tests
+
+# List all unit test suites and the names of the tests they contain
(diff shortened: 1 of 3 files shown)


topics/testing/README.md
```diff
@@ -1,6 +1,21 @@
 # Testing
 Before proceeding ensure the tests were built using the [build procedures].
 
+Both unit and integration tests use [Boost.Test](http://www.boost.org/doc/libs/1_62_0/libs/test/doc/html/index.html)
```
mikefero: +1 on the description; however should we also indicate how to launch them in Windows?

topics/testing/README.md
```diff
@@ -1,6 +1,21 @@
...
+# Run all integration tests
+test/integration_tests/cassandra_integration_tests
+
+# List all unit test suites and the names of the tests they contain
+test/unit_tests/cassandra_unit_tests --list_content
+
+# Run a particular suite
+test/unit_tests/cassandra_unit_tests -t blacklist_lb
+```
+
```
mikefero: - There is also a configuration file (that is being deprecated) in favor of command line driven options, we might also want to include information about the [config.txt](https://github.com/datastax/cpp-driver/blob/master/test/ccm_bridge/data/config.txt) file until it is fully removed and that it should be located in the working directory (integration tests only as the unit tests do not require any additional configuration elements to run). I don't think we should document the defaults as we update them during release to use the latest Cassandra version at the time.
- CCM information is also missing as a dependency (this is required for the integration tests to run properly)

PR #330, opened 2017-01-04
418 - Add support for Duration datatype

stamhankar999:


src/encode.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+namespace cass {
+
+static Buffer encode_internal(CassDuration value, bool with_length) {
+  // Each duration attribute needs to be converted to zigzag form. Use an array to make it
+  // easy to do the same encoding ops on all three values.
+  uint64_t zigzag_values[3];
+
+  // We need varint sizes for each attribute.
+  size_t varint_sizes[3];
```
mpenick: What do you think about changing the name of these to  "vint" instead of "varint"? "varint" is already a Cassandra type.

src/utils.hpp
```diff
@@ -91,6 +91,52 @@ std::string& to_cql_id(std::string& str);
...
+}
+
+inline cass_uint64_t encode_zig_zag(cass_int64_t n) {
+  return (n << 1) ^ (n >> 63);
+}
+
+inline size_t varint_size(cass_int64_t value) {
+  // | with 1 to ensure magnitude <= 63, so (63 - 1) / 7 <= 8
+  size_t magnitude = num_leading_zeros(value | 1);
+  return magnitude ? (9 - ((magnitude - 1) / 7)) : 9;
```
mpenick: This doesn't need a branch: (0 - 1) / 7 == 0

src/encode.cpp
```diff
@@ -0,0 +1,84 @@
...
+    if (varint_sizes[i] == 1) {
+      // This is just a one byte value; write it and move on.
+      *cur_byte++ = zigzag_values[i];
+      continue;
+    }
+
+    // Write the bytes of zigzag value to the data array with most significant byte
+    // in first byte of buffer (cur_byte), and so on.
+    for (int j = varint_sizes[i] - 1 ; j >= 0 ; --j) {
+      *(cur_byte + j) = zigzag_values[i] & 0xff;
```
mpenick: Should this be logic be encapsulated in an `char* encode_vint(char* output, int64_t value, size_t)` function?

src/utils.hpp
```diff
@@ -91,6 +91,52 @@ std::string& to_cql_id(std::string& str);
...
+  // index is the (zero based) index, counting from lsb, of the most-significant 1 bit.
+  // For example, a value of 12 (b1100) would return 3. The 4th bit is set, so there are
+  // 60 leading zeros.
+  return 64 - index - 1;
+#else
+  return __builtin_clzll(value);
+#endif
+}
+
+inline cass_int64_t decode_zig_zag(cass_uint64_t n) {
```
mpenick: `decode_zig_zag()` an `encode_zig_zag()` might be a better fit in `serialization.hpp`?

src/value.cpp
```diff
@@ -139,6 +139,71 @@ CassError cass_value_get_bytes(const CassValue* value,
...
+  for (ctr = 0; ctr < 3 && cur_byte != end; ++ctr) {
+    int num_extra_bytes;
+    int i;
+    cass_byte_t first_byte = *cur_byte++;
+    if (first_byte <= 127) {
+      // If this is a multibyte vint, at least the MSB of the first byte
+      // will be set. Since that's not the case, this is a one-byte value.
+      *outs[ctr] = first_byte;
+    } else {
+      // The number of consecutive most significant bits of the first-byte tell us how
```
mpenick: Should this logic also be encapsulated in a function `char * decode_vint(char* input, int64_t* out)`?

src/encode.cpp
mpenick: +1 w/ comments include and the custom type change we talked about. Good work.

mpenick:
+1 w/ comments include and the custom type change we talked about. Good work.

src/encode.cpp
stamhankar999: Sounds good; will do.

src/utils.hpp
stamhankar999: magnitude is an unsigned type, so subtracting 1 and then dividing by 7 causes it to be junk. We could cast magnitude to an int when doing the calculation instead of the branch...

src/encode.cpp
stamhankar999: Ok, sounds good.

src/utils.hpp
stamhankar999: ahh, didn't realize that was there. Will move it.

src/value.cpp
stamhankar999: Sounds good, will do.

src/utils.hpp
mpenick: Ah, make sense.

src/result_response.cpp
```diff
@@ -195,7 +210,8 @@ class DataTypeDecoder {
 
 private:
   char* buffer_;
-  DataType::Ptr data_type_cache_[CASS_VALUE_TYPE_LAST_ENTRY];
+  sparsehash::dense_hash_map<uint16_t, DataType::Ptr> data_type_cache_;
```
mpenick: I'd like to move this back to using an array now that the potential values are more dense.

PR #331, opened 2017-01-16
354 - Add pid to node hash

stamhankar999:


mpenick:
+1 Please squash before merging into master.

PR #332, opened 2017-01-18
356 - Map recognized custom types to built-in types

stamhankar999:


src/result_response.cpp
```diff
@@ -104,7 +104,9 @@ namespace cass {
 class DataTypeDecoder {
 public:
   DataTypeDecoder(char* input)
-    : buffer_(input) { }
+    : buffer_(input) {
+    native_types_.init_class_names();
```
mpenick: This is very expensive and would have to be initialized for each type decode for every result set.

PR #334, opened 2017-01-20
428 - Use global primitive data types instead of local data type caches

mpenick:


src/generate_value_types.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+# 'register' is a deprecated keyword in C++1x
+gperf -t value_types_by_class.gperf | sed 's/register //g' > value_types_by_class.hpp
+gperf -t value_types_by_cql.gperf | sed 's/register //g' > value_types_by_cql.hpp
```
stamhankar999: If a user calls the script from outside of the src dir, the generated files won't go to the src directory. How about something like this (untested):
```
DIRNAME=`dirname "$0"`
for i in value_types_by_class value_types_by_cql ; do
  gperf -t "$DIRNAME/$i.gperf" | sed 's/register //g' > "$DIRNAME/$i.hpp"
done
```
No one should be insane enough to use spaces in their directory paths, but the extra quoting above handles it if they do.

stamhankar999:
This looks great! My only gripe is that when adding a new type to the system, we have to update three files: the CASS_VALUE_TYPE_MAPPING in cassandra.h and the two .gperf files. The former actually has some of the same info as the value_types_by_cql.gperf file.

Consider this alternative:
* Add the class-names in CASS_VALUE_TYPE_MAPPING
* Just as we have the DataType::native_types_ class variable, add a map variable for mapping class-names to DataType's (or the enum value), and add a map variable for mapping cql-type-names to DataType's.
* Have the DataTypeInitializer initialize the maps

If it's an STL map, we'll have log(n) access, which is, of course, worse than the O(1) hash access provided by gperf. If that's a concern, maybe there's a general hash-map impl we can leverage. It may not be perfect hashing, but I think that would be good enough.

src/data_type.cpp
```diff
@@ -383,68 +385,39 @@ namespace cass {
...
-  DataTypeMap::const_iterator i = by_cql_names_.find(name);
-  if (i == by_cql_names_.end()) return DataType::NIL;
-  return i->second;
+DataType DataType::native_types_[CASS_VALUE_TYPE_LAST_ENTRY];
+
+struct DataTypeInitializer {
+  DataTypeInitializer() {
+    // Add a reference so that the memory is never deleted
+#define XX_VALUE_TYPE(name, type, cql) (new(&DataType::native_types_[name]) DataType(name))->inc_ref();
+    CASS_VALUE_TYPE_MAPPING(XX_VALUE_TYPE)
```
mpenick: This doesn't allocate memory. This is a "placement new".

src/data_type.cpp
```diff
@@ -383,68 +385,39 @@ namespace cass {
...
+  }
+};
+
+DataTypeInitializer __data_type_initializer__;
+
+DataType::ConstPtr DataType::get_native_by_class(const std::string& name) {
+  ValueTypeByClassMapping* mapping =
+    ValueTypeByClass::in_word_set(name.c_str(), name.length());
+  if (mapping == NULL) return DataType::NIL;
+  return DataType::ConstPtr(&native_types_[mapping->value_type]);
```
mpenick: The memory is statically allocated and initialized with "placement new". There's nothing to deallocated.

mpenick:
I'm not concerned O(log(n)) vs O(1). I'm trying to avoid heap allocation from being done in the static context. We did this a bunch when we first release the driver and had clients that had issues. I'm trying to remember what the issues were (I'll get back to you when I remember or find the issues). At first I wasn't even going to use static initialization (because this can have issues too). I was going to use a pattern similar to SSL's initialization, but this seems to be simple enough of a use case and doesn't have any heap allocation.

mpenick:
Side note: We could have moved the `std::map` to the `dense_hash_map` for O(1) performance. This is something I highly recommending with code moving forward. Esp. in cases where you can find a suitable empty key and/or deleted key.

mpenick:
I remembered the issue. Allocating heap memory in static initialization prevents users from allocating that memory using their custom heap allocator. Static initialization usually happens when the library is loaded, before the client application has had a chance to run, so the client application is unable to set their allocator for memory allocated during that prior static initialization.

Custom allocation is something we plan to support, in short order, because it has been requested by several major users. The issue: https://datastax-oss.atlassian.net/browse/CPP-360

mpenick:
I agree, it's not great having that same information duplicated in few files.  Maybe those .gperf files can be generated on-the-fly, as an intermediate step, using the information from`cassandra.h`?

src/generate_value_types.sh
mikefero: I disagree with the no spaces in directory paths; Windows still loves spaces and a good number of those users still utilizes spaces in their paths especially when their user directory is their fullname which may contain spaces.

These are my concerns:

1. Using a shell script to generate these files requires that we also make a batch file for Windows; would prefer that we utilize CMake to handle the generation instead
2. Committing generated files could lead to a situation where we have forgotten to manually generate the files and commit; granted chances are low
3. Not making this an automated step during the build process turns this into a manual process when we should probably automate it

2) and 3) are kind of the same and I am not 100% against keeping this generation manual and committing the files when required as the amount of times these will change is extremely low.

src/data_type.cpp
stamhankar999: Yeah, I get that this is "placement new", but I don't understand why the inc_ref is needed.

mpenick:
Heh, Since you both thumbs up it do you want to write the generator? :)

src/data_type.cpp
mpenick: It avoids ever call `delete` on the statically allocated memory.

src/data_type.cpp
mpenick: yes

src/data_type.cpp
```diff
@@ -383,68 +387,48 @@ namespace cass {
...
+DataType::ConstPtr SimpleDataTypeCache::by_cql(StringRef name) {
+  ValueTypeByCqlMapping* mapping =
+    ValueTypeByCql::in_word_set(name.data(), name.length());
+  if (mapping == NULL) return DataType::NIL;
+  return DataType::ConstPtr(get_for_thread(mapping->value_type));
+}
+
+DataType::ConstPtr SimpleDataTypeCache::by_value_type(uint16_t value_type) {
+  if (value_type == CASS_VALUE_TYPE_UNKNOWN ||
+      value_type <= CASS_VALUE_TYPE_LAST_ENTRY) {
```
stamhankar999: Shouldn't this be >=?

src/data_type.hpp
```diff
@@ -452,18 +432,33 @@ class UserType : public DataType {
...
+// threads simultaneously modify the reference count of shared simple
+// data type instances.
+#define COPIES_PER_SIMPLE_DATA_TYPE 5
+
+// Allocations have been a bottleneck when handling response metadata.
+// This class initializes and retrieves simple data types (e.g. varchar, int
+// etc) so these data types don't need to be allocated/deallocated over and over.
+// `DataType` is reference counted so it could lead to mulitple threads modifying
+// a shared reference count. To mitigate this sharing, thread IDs are used
+// to distribute mulitple copies of the same data type.
```
stamhankar999: mulitple => multiple

Love the comment blocks here and above.

mpenick:
This is finally ready to review again. This iteration doesn't require having value type information in multiple places and is simpler. This uses `dense_hash_map` with a static allocator so its a hybrid between the two approaches.

src/small_dense_hash_map.hpp
```diff
@@ -0,0 +1,72 @@
...
+    : public sparsehash::dense_hash_map<
+      K, V,  HashFcn, EqualKey,
+      FixedAllocator<std::pair<const K, V>, MIN_BUCKETS(N)> > {
+public:
+  typedef std::pair<const K, V> Pair;
+  typedef FixedAllocator<std::pair<const K, V>, MIN_BUCKETS(N)> Allocator;
+  typedef sparsehash::dense_hash_map<K, V, HashFcn, EqualKey, Allocator> HashMap;
+
+  SmallDenseHashMap()
+    : HashMap(N, typename HashMap::hasher(), typename HashMap::key_equal(), Allocator(&fixed_)) {
```
stamhankar999: Why use `typename` here but not in the other constructor? Also, aren't these actual values, not types, so we shouldn't have `typename` here?

test/unit_tests/src/test_data_type.cpp
```diff
@@ -382,6 +383,66 @@ BOOST_AUTO_TEST_CASE(subtypes)
   }
 }
 
+BOOST_AUTO_TEST_CASE(value_types_by_class)
+{
+#define XX_VALUE_TYPE(name, type, cql, klass)             \
+  BOOST_CHECK(strlen(klass) == 0 ||                       \
+              cass::ValueTypes::by_class(klass) == name); \
+
+  CASS_VALUE_TYPE_MAPPING(XX_VALUE_TYPE)
```
stamhankar999: Very neat use of the macro!

src/data_type.cpp
```diff
@@ -383,68 +384,73 @@ namespace cass {
...
+      value_type == CASS_VALUE_TYPE_MAP ||
+      value_type == CASS_VALUE_TYPE_SET ||
+      value_type == CASS_VALUE_TYPE_UDT ||
+      value_type == CASS_VALUE_TYPE_TUPLE ||
+      value_type >= CASS_VALUE_TYPE_LAST_ENTRY) {
+    return DataType::NIL;
+  }
+  DataType::ConstPtr& data_type = cache_[value_type];
+  if (!data_type) {
+    data_type = DataType::ConstPtr(
```
mpenick: It's a reference.

src/small_dense_hash_map.hpp
mpenick: Fixing compiler errors. Sometimes C++ compiler get confused and need told that a name refers to a type.  Perhaps it should be added to both, and the second constructor didn't error because it's not being currently used?

src/data_type.hpp
```diff
@@ -452,18 +435,37 @@ class UserType : public DataType {
...
+                            StringRefIHash,
+                            StringRefIEquals> HashMap;
+
+  static HashMap value_types_by_class_;
+  static HashMap value_types_by_cql_;
+};
+
+class SimpleDataTypeCache {
+public:
+  const DataType::ConstPtr& by_class(StringRef name) {
```
mpenick: I really concerned about sharing reference counted objects on multiple threads. With this method we get some sharing without causing multiple cores to thrash on those shared reference counts.

src/small_dense_hash_map.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include "macros.hpp"
+
+#include <assert.h>
+#include <sparsehash/dense_hash_map>
+
+// This should be the same as sparsehash::densehashtable<>::HT_OCCUPANCY_PCT
+// which can be found at the bottom of "sparesehash/internal/densehashtable.h".
+#define OCCUPANCY_PCT 50
+
+#define MIN_BUCKETS(N) STATIC_NEXT_POW_2((((N * 100) / OCCUPANCY_PCT) + 1))
```
mpenick: This is to statically allocate enough buckets to hold N items without requiring additional allocations. A hash table doesn't just allocate N buckets to hold N items because of hash collisions so it allocates a number of buckets over the current number of items (controlled by a load factor) to compensate for that.

`OCCUPANCY_PCT` is `dense_hash_map`'s load factor and it needs the number of buckets to be a power of two because it uses  a mask and `&` to wrap around instead of using modulo (which is expensive).

This calculates the number of buckets considering the load factor: `(N * 100) / OCCUPANCY_PCT ` (The `+ 1` is to allocate at least one bucket) then it rounds up to the next power of two for the wrap around trick.

src/small_dense_hash_map.hpp
mpenick: Also, this is evaluated as `(typename HashMap::hasher)()` not `typename (HashMap::hasher())`.

src/data_type.cpp
stamhankar999: Oh, duh! Makes sense.

src/data_type.hpp
stamhankar999: Do we need reference counting for this? What if it's a singleton initialized at the beginning of time and consumers get the elements they need? The consumers don't even try to delete what they've got, and when the application terminates, the singleton will be destroyed then. by_class and similar methods could return a `const DataType&` or `const DataType*`; the former would make it less likely that a consumer would try to destroy the DataType.

src/small_dense_hash_map.hpp
stamhankar999: Coolness. Makes sense. They key thing I was missing was that the wrap around is done by mask and requires num-buckets to be a power of 2.

src/small_dense_hash_map.hpp
stamhankar999: Ohhh... that does make more sense.

src/data_type.hpp
mpenick: The problem is the external API and other places in the code treat more complex and unique data types (collections, UDTs, tuples, etc.) as generic `DataType`.

mpenick:
Merged at 3c55e33328dd36c78b43be754a5056e7ddc5b6ea

PR #335, opened 2017-01-23
412 - Make timestamp generator monotonic

mpenick:
* Timestamps now use microsecond precision instead of millisecond.


mpenick:
It looks like I need to fix the format specifiers in that clock skew warning log message. I'll do that tomorrow.

include/cassandra.h
```diff
@@ -9477,6 +9477,12 @@ cass_timestamp_gen_server_side_new();
  * exceeded then a warning is logged and timestamps stop incrementing until the next
```
stamhankar999: The above comment-block probably needs to be updated -- the new code has proper microsecond precision and bumps the timestamp by 1 when there is clock skew, rather than using the lowest 3 digits as a counter?

include/cassandra.h
```diff
@@ -9485,12 +9491,27 @@ cass_timestamp_gen_server_side_new();
...
  */
 CASS_EXPORT CassTimestampGen*
 cass_timestamp_gen_monotonic_new();
 
 /**
+ * Same as cass_timestamp_gen_monotonic_new(), but with settings for controlling
+ * warnings about clock skew.
+ *
+ * @param warning_threshold_us The amount of clock skew, in microseconds, that
+ *                             must be detected before a warning is triggered.
```
stamhankar999: You should mention that setting this < 0 will disable warnings generation.

src/timestamp_generator.cpp
```diff
@@ -52,24 +61,31 @@ int64_t MonotonicTimestampGenerator::next() {
...
-      ++counter;
+  int64_t current = get_time_since_epoch_us();
+
+  if (last >= current) { // There's clock skew
+    // If we execeed our warning threshhold then warn periodically that clock
+    // skew has been detected.
+    if (warning_threshold_us_ >= 0 && last > current + warning_threshold_us_) {
+      int64_t now = get_time_since_epoch_ms();
+      int64_t last_warning = last_warning_.load();
+      if (now > last_warning + warning_interval_ms_ &&
```
stamhankar999: With the default settings, will this trigger a warning after the first? If your clock is skewed by 1 sec (as determined  in line 72), your `now` will be 1 sec less than real current time. If your last warning was emitted a second ago, then `now` won't be greater than `last_warning + warning_interval_ms_`. OTOH, last_warning was recorded with a skewed time, so maybe it balances out.

Maybe there isn't an issue here, but worth mulling over some more to be sure.

src/timestamp_generator.hpp
```diff
@@ -61,16 +61,24 @@ class ServerSideTimestampGenerator : public TimestampGenerator {
...
   virtual int64_t next();
 
 private:
   int64_t compute_next(int64_t last);
 
   Atomic<int64_t> last_;
+  Atomic<int64_t> last_warning_;
+
+  const int64_t warning_threshold_us_;
+  const int64_t warning_interval_ms_;
```
stamhankar999: I understand that `warning_threshold_us_` is signed so that a user can set it negative to indicate "don't emit warnings"; why is `warning_interval_ms_` a signed 64-bit int?

stamhankar999: If either of them being negative indicates disabling skew warnings, that's cool, but it should be documented as such and line 72 should check that `warning_threshold_ms_` is non-negative.

test/unit_tests/src/test_timestamp_gen.cpp
```diff
@@ -18,31 +18,114 @@
...
+    BOOST_CHECK(now > prev);
+    prev = now;
+  }
+  double elapsed = (cass::get_time_since_epoch_ms() - start) / 1000.0; // Convert to seconds
+
+  // We can generate at most 1,000,000 timestamps in a second. If we exceed this
+  // limit and the clock skew threshold then a log should have been printed.
+  double timestamp_rate = static_cast<double>(NUM_TIMESTAMPS) / elapsed;
+  if (timestamp_rate > 1000000.0 && elapsed > 1.0) {
+    BOOST_CHECK(counter > 0);
```
stamhankar999: In practice, does the test get here? Does it take at least 1.0 seconds to get 5,000,000 timestamps?

test/unit_tests/src/test_timestamp_gen.cpp
```diff
@@ -18,31 +18,114 @@
...
+  if (timestamp_rate_100ms > 1000000.0 && elapsed_100ms > 1.0) {
+    BOOST_CHECK(counter_100ms > 0);
+  } else {
+    BOOST_TEST_MESSAGE("Warning: The 100ms generator did not exceed its maximum rate.");
+    BOOST_CHECK(counter_100ms == 0);
+  }
+
+  // The 100ms timestamp generator should have logged more times because
+  // it had a shorter iterval.
+  BOOST_CHECK(counter_100ms > counter_default);
```
stamhankar999: I was confused at first that you're not doing any magic to fake out the clock (to make actual clock skew) but you get a good approximation by getting the current-time multiple times per microsecond and making sure the timestamp is still monotonically increasing. Good stuff.

src/timestamp_generator.cpp
mpenick: Good catch. This is interesting. I guess this could have if your clock skews by another second consecutively several times in a row. I'm think about this.

include/cassandra.h
mpenick: Good catch. Fixing.

test/unit_tests/src/test_timestamp_gen.cpp
mpenick: Because the default logging threshold is 1 second of skew. If the tests finished before 1 second a warning won't occur.

src/timestamp_generator.cpp
mambocab: We just ran into this decision too. We decided that either behavior is reasonable, so we chose not to account for the 1-second offset to match the Java driver's behavior, as we understand it:

https://github.com/datastax/python-driver/pull/681#discussion_r97587655

Open to hearing other opinions if people have strong ones, but I don't see a strong reason to diverge from the Java driver.

test/unit_tests/src/test_timestamp_gen.cpp
mpenick: Yes. I don't expect the warning to trigger if with the default settings unless we sustain a timestamp generation rate greater than one per millisecond for at least one second. Maybe this is incorrect?

mpenick: Hmmm. I think a better test might use time instead of an arbitrary number of timestamps...I'm fixing this.

src/timestamp_generator.hpp
mpenick: I changed this to make `warning_threshold_ms_` zero if it's a negative number. I also added documentation. What do you think?

src/timestamp_generator.cpp
mpenick: A real fix is to use a monotonic system clock to manage the warning interval. The problem is getting that to work properly on all supported platforms (esp. in Java and Python). It might be overkill for handling the warnings, but I'm going at least investigate how hard it would be to implement a millisecond precision monotonic clock for the C++ driver's major platforms (BSD, Linux, Windows).

src/get_time.cpp
```diff
@@ -31,32 +33,121 @@ namespace cass {
...
+
+uint64_t get_time_monotonic_ns() {
+  if (ClockInfo::supports_monotonic()) {
+    struct timespec tp;
+    clock_gettime(CLOCK_MONOTONIC, &tp);
+    return static_cast<uint64_t>(tp.tv_sec) * NANOSECONDS_PER_SECOND +
+           static_cast<uint64_t>(tp.tv_nsec);
+  } else {
+    return get_time_since_epoch_us() * NANOSECONDS_PER_MICROSECOND;
+  }
```
stamhankar999: I'm torn on one point here -- this whole file is platform-specific, so I feel like it would be better to split it into three files (say get_time-win.cpp, get_time-mac.cpp, get_time-unix.cpp) to make it easier for a developer to look at in the future. On the other hand, it's 150+ lines of code, which isn't that big.

Consider splitting, but I'm fine either way.

src/timestamp_generator.hpp
stamhankar999: You mean `warning_interval_ms_`. But why not making it a uint64_t then? A user who supplies a negative value will get a compiler warning, wouldn't he?

mpenick: No, it doesn't work like that. Try compiling this:  https://gist.github.com/mpenick/7878d5d328bbb9aa74a1f7dd125ebe15

```
$ clang++ -Wall -Wextra -pedantic unsigned.cpp
unsigned.cpp:5:30: warning: 'long long' is an extension when C99 mode is not enabled [-pedantic,-Wlong-long]
  printf("%llu\n", (unsigned long long)value);
                             ^
1 warning generated.
```

A warning is not generated for the negative value.

mpenick: Yes, I meant `warning_interval_ms`. 

I think it was a mistake to have unsigned values exposed in the API. I've been attempting to move away from it.

src/get_time.cpp
mpenick: Good call. I didn't realized this file was getting big.

src/timestamp_generator.hpp
stamhankar999: Interesting that the compiler doesn't warn when it knows you're trying to put a negative value into an unsigned type.

Also, why do you think it was a mistake to expose unsigned values? It shows an intent for valid argument values (even if the compiler doesn't enforce it).

mpenick: > It shows an intent for valid argument values.

I used to think this too, but for the pitfall shown in the above example it's not worth it. If a user passes a negative number the value turns into a really large number and it's not possible to detect an error. It's better to used signed values and return an explicit error (or at least document and correct it).

stamhankar999: Good point. Signed it is! :)

mpenick:
Squashed and merged at 16577c504ea65dddccea56723c09c6cbbfb25370

PR #337, opened 2017-01-27
Fixes and cleanup found while reviewing code

mpenick:


src/utils.cpp
```diff
@@ -23,7 +23,7 @@
 #include <functional>
 #include <sstream>
 
-#if defined(_MSC_VER)
+#if (defined(WIN32) || defined(_WIN32))
```
stamhankar999: Is WIN32 defined for 64-bit Windows? We need windows.h for both.

stamhankar999:
Other than my question about 64-bit Windows, lgtm.

mpenick:
`WIN32` means the Windows API. It's always defined regardless of the architecture: https://msdn.microsoft.com/en-us/library/b0084kay.aspx

stamhankar999:
👍  Merge at will.

mpenick:
Merged at 8e7b256023d40830730e1cb24978f5a9b658d064

PR #339, opened 2017-01-30
398 - Invalid local DC can cause connect hang

mpenick:
Fix for an issue where a session can hang during its connection process
because there are no hosts available in the local DC.


stamhankar999:
👍  LGTM.

PR #340, opened 2017-01-31
test: 398 - Invalid local DC causing driver hang

mikefero:


test/integration_tests/src/test_control_connection.cpp
```diff
@@ -465,4 +465,43 @@ BOOST_AUTO_TEST_CASE(randomized_contact_points)
...
+ *
+ * @since 2.6.0
+ * @test_category control_connection
+ * @jira_ticket CPP-398
+ * @expected_result Driver will not hang and session/control connection will
+ *                  terminate; CASS_ERROR_LIB_NO_HOSTS_AVAILABLE
+ */
+BOOST_AUTO_TEST_CASE(invalid_dc)
+{
+  // Create the CCM cluster if it does not already exist (2 DCs)
```
stamhankar999: Do you need 2 DCs for this test? It seems to me you just need the one (which will be considered remote since "invalid_dc" is considered local).

mikefero: Theoretically no, but I wanted to re-create a pseudo realistic scenario since it involves data centers.

mpenick:
+1

mikefero:
Merged at 78497ff

PR #341, opened 2017-02-01
test: 408 - Reconnection if SSL enabled on crashed and restarted server

mikefero:
- Updated CCM bridge to send SIGHUP to cluster/node(s) instance


mpenick:
+1

mikefero:
Merged at 34674fc

PR #343, opened 2017-02-01
429 - Update Duration tests and example to not mix positive and negat…

stamhankar999:
…ive attributes (which is illegal).

### DIFF
--- a/examples/duration/duration.c
+++ b/examples/duration/duration.c
@@ -170,13 +170,11 @@ int main(int argc, char* argv[]) {
 
   insert_into(session, "base", 0, 0, 0);
   insert_into(session, "simple", 1, 2, 3);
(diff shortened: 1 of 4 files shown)


mikefero:
These values when inserted into the server were not throwing server-side errors; is this a bug on the server and should a ticket be created to produce errors? If that is the case then we should write another test where we ensure the server-side errors for the duration type are being picked up properly by the driver.

stamhankar999:
C* was not doing server-side validation initially, but Ben Lerer recently added it:

https://issues.apache.org/jira/browse/CASSANDRA-13143

The fix-version on that ticket says 3.10, but the comments indicate it is actually resolved on 3.11.

mpenick:
+1

PR #346, opened 2017-02-13
CPP-433 - Request metrics returning zero percentiles

mpenick:
Histogram metrics were incorrectly counting the same values repeatedly.


stamhankar999:
👍

PR #347, opened 2017-02-15
CPP-434

stamhankar999:
Update api, tests, docs, examples to use 32-bit int's for Duration attributes.

### DIFF
--- a/src/value.cpp
+++ b/src/value.cpp
@@ -179,9 +179,9 @@ static const cass_byte_t* decode_vint(const cass_byte_t* input, const cass_byte_
   return input;
 }
 
-CassError cass_value_get_duration(const CassValue* value, cass_int64_t* months, cass_int64_t* days, cass_int64_t* nanos)
+CassError cass_value_get_duration(const CassValue* value, cass_int32_t* months, cass_int32_t* days, cass_int32_t* nanos)
 {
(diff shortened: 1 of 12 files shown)


mpenick:
`cass::CassDuration` is still using `int64_t`: https://github.com/datastax/cpp-driver/blob/cpp-434/src/types.hpp#L65-L75

mpenick:
Other than the above issue +1

src/value.cpp
```diff
@@ -199,10 +199,11 @@ CassError cass_value_get_duration(const CassValue* value, cass_int64_t* months,
   end = cur_byte + data_size;
 
   for (ctr = 0; ctr < 3 && cur_byte != end; ++ctr) {
-    cur_byte = decode_vint(cur_byte, end, outs[ctr]);
+    cass_int64_t decoded = 0;
+    cur_byte = decode_vint(cur_byte, end, &decoded);
     if (cur_byte == NULL)
       return CASS_ERROR_LIB_BAD_PARAMS;
-    *outs[ctr] = cass::decode_zig_zag(*outs[ctr]);
+    *outs[ctr] = (cass_int32_t) cass::decode_zig_zag(decoded);
```
mpenick: Don't use C style casts. Use `static_cast<>` here

PR #351, opened 2017-03-13
Add missing include for MSVC15

SeverinLeonhardt:
The _BitScanReverse functions now require intrin.h to be included.


mpenick:
Thanks.

PR #353, opened 2017-04-15
rebuild token map on new node

cangeli-orange:
I know this solution is not acceptable, but I noticed that rebuilding the tokens map instead of updating it was faster and less CPU consuming when adding a node to a cluster of more than 40 nodes and 10 keyspaces.


mpenick:
Interesting, thanks for the patch. I'll investigate why updating is slower...

mpenick:
I found a bug that fixes the issue of updating vs rebuilding. Patch incoming...

mpenick:
Patch: https://github.com/datastax/cpp-driver/pull/357
JIRA: https://datastax-oss.atlassian.net/browse/CPP-464

PR #355, opened 2017-04-30
Hide all symbols except explicitly exported with CASS_EXPORT.

API92:
It fixes crash (or assert failure in Debug configuration) when some .so uses other version (1.1) of rapidjson from /usr/include and overrides some (not all) rapidjson methods compiled in libcassandra.so. This reproduced when connecting to scylladb. Under debugging it looks like rapidjson::Document::StartObject (from /usr/include/rapidjson/...) called with wrong this pointer from GenericParser::ParseObject (from third_party/rapidjson/...), but preprocessor output shows only sources from third_party/rapidjson.


mpenick:
Thanks for the patch. I'll take a look at this sometime this week.

src/macros.hpp
```diff
@@ -77,4 +77,20 @@ struct StaticNextPow2 {
   enum { value = StaticNextPow2Helper<8 * sizeof(size_t) - 1, N>::value };
 };
 
+#if !defined(CASS_STATIC)
+#  if (defined(WIN32) || defined(_WIN32))
+#    if defined(CASS_BUILDING)
+#      define CASS_IMPL_EXPORT __declspec(dllexport)
+#    else
+#      define CASS_IMPL_EXPORT __declspec(dllexport)
+#    endif
```
stamhankar999: The if and else bodies are identical. Why have this if for CASS_BUILDING?

API92: This code copied from include/cassandra.h. In commit https://github.com/datastax/cpp-driver/commit/fb4c1a6f4c8824bc5198c7f404609e3a8d22a997 the message says that it solves some Windows warnings. But I can rewrite this code with single #define.

src/macros.hpp
```diff
@@ -77,4 +77,20 @@ struct StaticNextPow2 {
   enum { value = StaticNextPow2Helper<8 * sizeof(size_t) - 1, N>::value };
 };
 
+#if !defined(CASS_STATIC)
+#  if (defined(WIN32) || defined(_WIN32))
+#    if defined(CASS_BUILDING)
+#      define CASS_IMPL_EXPORT __declspec(dllexport)
+#    else
+#      define CASS_IMPL_EXPORT __declspec(dllexport)
```
mpenick: I'm pretty sure the `#else` should be `__declspec(dllimport)`. This is an existing typo. Let me do some research.

mpenick:
There was an issue where `dllexport` was on both sides of the `CASS_BUILDING` branch. Please open a new PR or a JIRA issue if this hasn't fixed your issue.

PR #359, opened 2017-06-06
There is no cass_schema_free() function; it's cass_schema_meta_free().

Castaglia:


include/cassandra.h
```diff
@@ -1988,7 +1988,7 @@ cass_session_get_schema_meta(const CassSession* session);
  * @param[in] session
  * @param[out] output
  *
- * @see cass_schema_free()
+ * @see cass_schema_meta_free()
```
stamhankar999: Actually, cass_session_get_metrics has nothing to do with cass_schema_meta_free; I suspect this is an old copy-paste error from above. Please remove this @see.

PR #360, opened 2017-06-06
Minor nit: Nagle's algorithm (not "Nagel's") is named after John Nagl…

Castaglia:
…e; see:

  https://en.wikipedia.org/wiki/Nagle%27s_algorithm


Castaglia:
This addresses [CPP-359](https://datastax-oss.atlassian.net/browse/CPP-359).

PR #361, opened 2017-06-23
Changing the randomisation generation :

devnexen:
- using getrandom syscall on linux, trying to read as we can in
non blocking mode, falling back to dev urandom eventually.
- on *BSD and MacOS, using arc4random_buf which does not fail.


mpenick:
@devnexen Thanks for taking the time to do this. I'll review this sometime this week.

src/random.cpp
```diff
@@ -84,33 +90,53 @@ uint64_t Random::next(uint64_t max) {
...
+    bool readurandom = true;
+#if defined(HAVE_GETRANDOM)
+    num_bytes = static_cast<ssize_t>(syscall(SYS_getrandom, &seed,
+            sizeof(seed), GRND_NONBLOCK));
+    if (num_bytes < static_cast<ssize_t>(sizeof(seed))) {
       char buf[STRERROR_BUFSIZE_];
       char* err = STRERROR_R_(errno, buf, sizeof(buf));
-      LOG_CRITICAL("Unable to open random device (%s): %s", device, err);
-      return seed;
+      LOG_WARN("Unable to read %zu random bytes (%s): %zu read",
```
mpenick: "%zu" isn't portable.

src/random.cpp
```diff
@@ -84,33 +90,53 @@ uint64_t Random::next(uint64_t max) {
...
+            sizeof(seed), err, num_bytes);
+    } else {
+      readurandom = false;      
     }
+#endif
```
mpenick: It would improve readability if this was `#endif // defined(HAVE_GETRANDOM)`. These nested `#if`s are hard to follow.

src/random.cpp
```diff
@@ -84,33 +90,53 @@ uint64_t Random::next(uint64_t max) {
...
+            static_cast<unsigned int>(sizeof(seed)),
+            static_cast<unsigned int>(num_bytes),
+            device, err);
+      }
+
+      close(fd);
     }
-
-    close(fd);
+#endif
```
mpenick: `#endif defined(HAVE_ARC4RANDOM)`

mpenick:
LGTM. Just a couple comments.

mpenick:
Thanks for the quick commit. Can you please rebase on master so we can get CI to pass?

PR #363, opened 2017-08-24
CPP325: Add the ability to set the default consistency level for statements

cormoran:
https://datastax-oss.atlassian.net/browse/CPP-325

- add api `cass_request_set_default_consistency`, ` cass_request_get_default_consistency` 
- initial default value is `CASS_CONSISTENCY_LOCAL_ONE` (unchanged)

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -1012,6 +1012,36 @@ cass_cluster_set_use_beta_protocol_version(CassCluster* cluster,
                                            cass_bool_t enable);
 
 /**
+ * Sets default consistency level of statement.
+ *
+ * <b>Default:</b> CASS_CONSISTENCY_LOCAL_ONE
+ *
+ * @public @memberof CassCluster
+ *
+ * @param[in] cluster
+ * @param[in] consistency
+ * @return CASS_OK if successful, otherwise an error occurred.
+ */
+CASS_EXPORT CassError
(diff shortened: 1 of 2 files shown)


mpenick:
Thanks for the pull request. As an external setting this should be added to the cluster object so that it can be configured per session.

cormoran:
If we add default consistency to Config, because of const reference to request object in session, we need to reflect them in RequestCallback. But, then, threre seems to be no way to pass correct consistency to DCAwarePolicy.

https://github.com/datastax/cpp-driver/blob/master/src/dc_aware_policy.cpp#L67

some solutions are .. 
to pass config object to 
- QueryPlan:: new_query_plan
- DCAwarePolicy Constructor

former seems better.
Is there any good solution ?

mpenick:
Passing the default consistency to `QueryPlan:: new_query_plan` wouldn't be too horrible.

cormoran:
Thank you for your advice.
I will work on it later in the week.

mpenick:
As of commit 5dc599f3a243c33a02d0c42edf4a80c8873c5008 this should be super easy to do per-session. The external methods just need to be added.

cormoran:
Very sorry for the latency -- failing to run integration test and personal matters kept me away form the project...

I updated the commit based on 5dc599f.

PR #364, opened 2017-08-24
Fix documentaion of installing openssl on mac

cormoran:
brew refuses `brew link --force openssl`(https://github.com/Homebrew/brew/pull/597). 
So we need to set openssl path in another way.  One way is to set `OPENSSL_ROOT_DIR`.

### DIFF
--- a/topics/building/README.md
+++ b/topics/building/README.md
@@ -72,8 +72,9 @@ additional configuration before building the driver.
 
 ```bash
 brew install openssl
-brew link --force openssl
+export OPENSSL_ROOT_DIR=/usr/local/opt/openssl
 ```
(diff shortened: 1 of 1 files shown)


mikefero:
@cormoran Thank you for your contribution. We have a different fix that will be applied in the upcoming release which [updates the OpenSSL hints](https://github.com/datastax/cpp-dse-driver/commit/449c457948ab720c2d4b9f31a7532604f17ecc93). This way the documentation can stay relatively generic for all the Mac OS flavors.

I am going to close this PR, but again thank you for bringing this to our attention so that we can get this into the next upcoming release of the driver.

mikefero:
@cormoran The fix has landed in [v2.7.1](https://github.com/datastax/cpp-driver/releases/tag/2.7.1) release of the driver.

PR #366, opened 2017-08-24
CPP-491: Unintentional switch case fallthrough in `RequestCallback`

mikefero:


mikefero:
NOTE: Failed TravisCI builds are due to a re-direction from `http` => `https` when downloading libuv; this has already been [corrected](https://github.com/datastax/cpp-driver/pull/362/files#diff-354f30a63fb0907d4ad57269548329e3R25) in #362.

PR #374, opened 2017-09-26
CPP:525 added a connection test in build.yaml

ArunTejCh:


build.yaml
```diff
@@ -69,6 +71,40 @@ release:
             ./build_${PACKAGE_TYPE}.sh
             find build -type f -name "*.${PACKAGE_TYPE}" -exec mv {} packages \;
             pushd packages
+
+            # Install cpp-driver packages
+            sudo ${PACKAGE_INSTALL} cassandra*.${PACKAGE_TYPE}
+
+            # Create test file against cpp-driver and try to connect. Should compile and throw a connection error when executed
+            $CC -x c -o $TEST_PROGRAM - -lcassandra - <<EOF
```
mpenick: You can remove the includes if you add `-Wno-implicit-function-declaration` to your compile command.

build.yaml
```diff
@@ -69,6 +71,40 @@ release:
...
+            #include <cassandra.h>
+            #include <stdio.h>
+
+            int main(int argc, char* argv[]) {
+            /* Setup and connect to cluster */
```
mpenick: These comment in the code are unnecessary for this.

build.yaml
mpenick: e.g. `echo 'int main() { cass_cluster_new(); printf("Hi\n"); }' | clang -x c - -Wno-implicit-function-declaration -lcassandra`

build.yaml
```diff
@@ -21,6 +21,8 @@ release:
       each:
         - script: |
             LIBUV_VERSION=1.14.1
+            CC=g++
```
mpenick: This can be `gcc` instead of `g++`.

mikefero: Should `CC` be an environment variable instead for the matrix (might be overkill though)?

build.yaml
```diff
@@ -69,6 +71,40 @@ release:
...
+            
+            # Throw error if either compilation or test run fails
+            if [[ $? -ne 0 ]] ; then
+              exit 1
+            fi    
+            
+            if [[ $($TEST_PROGRAM) != *"Success"* ]] ; then
+              exit 1
+            fi
+            rm test.*
```
mikefero: What is the `rm test.*` for?

build.yaml
```diff
@@ -69,6 +71,40 @@ release:
...
+            return 0;
+            }
+            EOF
+            
+            # Throw error if either compilation or test run fails
+            if [[ $? -ne 0 ]] ; then
+              exit 1
+            fi    
+            
+            if [[ $($TEST_PROGRAM) != *"Success"* ]] ; then
```
mikefero: * Why the different style bracketing for the conditional? Prefer consistency.

build.yaml
```diff
@@ -69,6 +71,40 @@ release:
...
+            connect_future = cass_session_connect(session, cluster);
+            cass_future_wait(connect_future);
+            printf("Success\n");
+            return 0;
+            }
+            EOF
+            
+            # Throw error if either compilation or test run fails
+            if [[ $? -ne 0 ]] ; then
+              exit 1
```
mikefero: `echo` an error message indicating that the compilation failed so that we can see that from the console log on Jenkins.

build.yaml
```diff
@@ -69,6 +71,40 @@ release:
...
+            fi    
+            
+            if [[ $($TEST_PROGRAM) != *"Success"* ]] ; then
+              exit 1
```
mikefero: Same here; `echo` to indicate the test program failed to run correctly.

build.yaml
ArunTejCh: I forgot to remove it.

build.yaml
mpenick: Double brackets is a bash thing that give you more advanced functionality.

mpenick: I agree. We should be consistent though.

build.yaml
ArunTejCh: All our snapshots currently use preinstalled gcc. We can change it to a matrix env if we enable CI and want to build against different compilers.

build.yaml
ArunTejCh: removed

build.yaml
ArunTejCh: Added

build.yaml
ArunTejCh: Looks like the string matching i'm using requires double brackets :(

build.yaml
```diff
@@ -69,6 +71,36 @@ release:
             ./build_${PACKAGE_TYPE}.sh
             find build -type f -name "*.${PACKAGE_TYPE}" -exec mv {} packages \;
             pushd packages
+
+            # Install cpp-driver packages
+            sudo ${PACKAGE_INSTALL} cassandra*.${PACKAGE_TYPE}
+
+            # Create test file against cpp-driver and try to connect. Should compile and throw a connection error when executed
```
mikefero: nit: Keep each line width to 100 max

PR #384, opened 2017-10-30
CPP-538 - Deprecate watermarks

mpenick:


src/session.cpp
```diff
@@ -135,7 +135,6 @@ void  cass_session_get_metrics(const CassSession* session,
 
 
   metrics->stats.total_connections = internal_metrics->total_connections.sum();
-  metrics->stats.available_connections = internal_metrics->available_connections.sum();
```
mikefero: We should probably set this to a value to; not sure if zero is the best for `available connection`, but I can't think of anything better if we are deprecating it and it can't be a negative value.

PR #386, opened 2017-11-01
CPP-471: Updates to README and copyright dates

mikefero:
This has been broken up into multiple commits for easier review; these commits will be squashed.


mpenick:
Everything looks good except I don't think it makes sense to get rid of "Inc", but we should add it where missing.

PR #388, opened 2017-11-08
CPP-543 - Fix dllimport/dllexport

mpenick:
Also, removed single build for static and shared libraries because they
use different compiler flags.


examples/async/CMakeLists.txt
```diff
@@ -12,3 +12,4 @@ add_dependencies(${PROJECT_EXAMPLE_NAME} ${PROJECT_LIB_NAME_TARGET})
 set_property(
   TARGET ${PROJECT_EXAMPLE_NAME}
   APPEND PROPERTY COMPILE_FLAGS ${CASS_EXAMPLE_C_FLAGS})
+set_property(TARGET ${PROJECT_EXAMPLE_NAME} PROPERTY FOLDER "Examples")
```
mikefero: 💯

PR #389, opened 2017-11-09
CPP-542/544: Port CPP-368 and CPP-432

mikefero:


mpenick:
There are additional driver fixed for CPP-368  in this commit: https://github.com/riptano/cpp-dse-driver/commit/f8fc72738e22a433d99e417077ca8bcf003fd998

mikefero:
I pulled those in #390 as I found out after the fact that I missed them (https://github.com/datastax/cpp-driver/pull/390/commits/68b7c0f9e65636910155926f63ce50f4895e5611). Would you prefer a rebase on both branches to get them into their respective PRs?

mpenick:
Nope. Just missed them in the other PR.

PR #397, opened 2017-12-18
CPP-539 - Adding RelWithDebugInfo build for Windows build

mikefero:
This update adds PDB files for all third party libraries and driver
binaries; they are also added to each archive during the packaging
phase.

* Defaulting build to RelWithDebugInfo
* Updated dependency versions for third party libraries
* Making driver PDB installation optional (defaults to true)


mikefero:
CPP-567 supersedes this; closing this pull request.

PR #400, opened 2018-01-15
Multiple fixes for documentation

alexott:
This includes:

- Fixing errors in examples
- Fixing incorrect descriptions
- Trying to improve readability of text
- Adding/fixing links to other documentation


topics/basics/handling_results/README.md
```diff
@@ -131,10 +131,10 @@ while (has_more_pages) {
   CassFuture* query_future = cass_session_execute(session, statement);
 
   const CassResult* result = cass_future_get_result(query_future);
+  cass_future_free(query_future);
```
mikefero: If the future is freed before the error is handled, then you will not be able to inspect the future to determine the type of error that may have occurred using `cass_future_error_result(query_future)`, `cass_future_error_code(query_future)`, or `cass_future_error_message(query_future)`. I think we should keep the `cass_future_free(query_future)` in the `if` block below the `/* Handle error */` section.

topics/basics/prepared_statements/README.md
```diff
@@ -1,11 +1,11 @@
 # Prepared Statements
 
-Prepared statements can be used to improve the performance of frequently executed queries. Preparing the query caches it on the Cassandra cluster and only needs to be performed once. Once created, prepared statements should be reused with different bind variables. Prepared queries use the `?` marked to denote bind variables in the query string.
+Prepared statements can be used to improve the performance of frequently executed queries. Preparing the query caches it on the Cassandra cluster and only needs to be performed once. Once created, prepared statements should be reused with different bind variables. Prepared queries use the `?` marker to denote "anonymous" bind variables in the query string. You can also use named bind variables if you specify them as `:name`.
```
mikefero: nit: "anonymous" => anonymous; I don't think the quotes are necessary here

topics/security/ssl/README.md
```diff
@@ -151,7 +152,7 @@ int load_trusted_cert_file(const char* file, CassSsl* ssl) {
 }
 ```
 
-It is possible to load multiple self-signed certificates or CA certificate chains. In the event where self-signed certificates with unique IP addresses are being used this will be required. It is possible to disable the certificate verification process, but it is not recommended.
+It is possible to load multiple self-signed certificates or CA certificate chains. This will be required in case when self-signed certificates with unique IP addresses are being used. It is possible to disable the certificate verification process, but it is not recommended.
```
mikefero: in `case` when => in `cases` when

topics/basics/data_types/README.md
```diff
@@ -1,9 +1,9 @@
 # Data Types
 
 [`CassDataType`] objects are useful for describing the different values that can
-be stored in Cassandra, from primitive types to more complex composite types
-such as UDTs, tuples and collections. Data types can be retrieved from existing
-metadata found in schema, results, values or prepared statements or they can be
+be stored in Cassandra, from primitive types to more complex composite types,
+such as, UDTs (user defined types), tuples and collections. Data types can be retrieved from existing
```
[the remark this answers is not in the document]
alexott: I did use this form, because UDT section in documentation is called "User Defined Types". I can update it as well...

topics/basics/handling_results/README.md
alexott: Ok, then I'll insert an additional `cass_future_free(query_future);` at the end of the loop

topics/basics/futures/README.md
```diff
@@ -4,7 +4,7 @@ Futures are returned from any driver call that would result in blocking the clie
 
 ## Waiting for Results
 
-Futures results can be waited on indefinitely by either calling a wait method or by attempting to get the result. They can also be waited on for an explicit amount of time or periodically polled without waiting to execute application code.
+Futures results can be waited on indefinitely by either calling the `cass_future_wait` method or by attempting to get the result with `cass_future_get_result`. They can also be waited on for an explicit amount of time or periodically polled without waiting to execute application code.
```
mpenick: `cass_future_wait` --> `cass_future_wait()`?
`cass_future_get_result` -- > `cass_future_get_result()`?

I find it easier to understand/read.

topics/basics/prepared_statements/README.md
```diff
@@ -1,11 +1,11 @@
...
 
 ```c
 /* Prepare the statement on the Cassandra cluster */
 CassFuture* prepare_future
-  = cass_session_prepare(session, "INSERT INTO example (key, value) VALUES (?, ?)");
+  = cass_session_prepare(session, "INSERT INTO example (key, value) VALUES (:key, :value)");
```
mpenick: Why this change? The values of the names can be determined by the `(key, value)` part of the query.

alexott: ok, will rollback - thought that it will be better visible...

topics/basics/tuples/README.md
```diff
@@ -29,7 +29,7 @@ cass_tuple_free(tuple);
 ## Create a Tuple using a Data Type
 
 A tuple can also be created using a [`CassDataType`] that comes from schema
-metadata or is manually constructed. However, this is not step necessary as
+metadata or is manually constructed. However, this is not necessary step as
```
mpenick: `this is not necessary step` --> `this is not **a** necessary step`

topics/basics/consistency/README.md
```diff
@@ -86,7 +86,7 @@ level can be found
   </tbody>
 </table>
 
-**NOTE:** Consistency level `CASS_CONSISTENCY_ANY` is only valid for read operation statements.
+**NOTE:** Consistency level `CASS_CONSISTENCY_ANY` is only valid for write operation statements.
```
mpenick: good catch!

topics/basics/prepared_statements/README.md
mpenick: I wouldn't call it anonymous at all. The names can often be determined by the query itself.

`SELECT * FROM test WHERE foo = ?`

The marker `?` can be determined by the name `foo` in the query string for both simple and prepared statements without having to explicitly name the marker. The applies to `INSERT` and `UPDATE` queires as well.

topics/basics/binding_parameters/README.md
```diff
@@ -13,7 +13,7 @@ cass_statement_bind_string(statement, 0, "abc");
 cass_statement_free(statement);
 ```
 
-Variables can only be bound by name for prepared statements. This limitation exists because query metadata provided by Cassandra is required to map the variable name to the variable’s marker index.
+Variables can only be bound by name for [prepared statements](../prepared_statements/). This limitation exists because query metadata provided by Cassandra is required to map the variable name to the variable’s marker index. Named bind variables are specified as `:variable` inside query string.
```
mpenick: This is no longer a limitation. This was just out-of-date. Simple statements now support named `?` parameters since the protocol v3 (or Cassandra 2.1).

So Cassandra 1.2/2.0 have this limitation and Cassandra 2.1+ does not.

mpenick:
Thanks for the updates. I did a quick pass on it.

topics/basics/uuids/README.md
```diff
@@ -63,5 +63,5 @@ cass_uint8_t version = cass_uuid_version(uuid);
 char uuid_str[CASS_UUID_STRING_LENGTH];
 cass_uuid_string(uuid, uuid_str);
 ```
-[`cass_uuid_timestamp()`]: http://datastax.github.io/cpp-driver/api/struct.CassUuid/#1a3980467a0bb6642054ecf37d49aebf1a
-[`CassUuidGen`]: http://datastax.github.io/cpp-driver/api/struct.CassUuidGen/
+[`cass_uuid_timestamp()`]: http://docs.datastax.com/en/developer/cpp-driver/latest/api/struct.CassUuid/#1a3980467a0bb6642054ecf37d49aebf1a
```
mikefero: Should be `http://docs.datastax.com/en/developer/cpp-driver/latest/api/struct.CassUuid/#function-cass_uuid_timestamp`

topics/basics/binding_parameters/README.md
alexott: Corrected it

topics/configuration/README.md
```diff
@@ -271,10 +271,10 @@ to wait for one or more nodes to respond back to the coordinator node before a
 request can complete. In multi-datacenter configurations, consistency levels such as
 `EACH_QUORUM` can cause a request to wait for replication across a slower cross
 datacenter network link.  More information about setting the consistency level
-can be found [here](http://datastax.github.io/cpp-driver/topics/basics/consistency/).
+can be found [here](http://docs.datastax.com/en/developer/cpp-driver/latest/topics/basics/consistency/).
 
-[`allow_remote_dcs_for_local_cl`]: http://datastax.github.io/cpp-driver/api/struct.CassCluster/#1a46b9816129aaa5ab61a1363489dccfd0
+[`allow_remote_dcs_for_local_cl`]: http://docs.datastax.com/en/developer/cpp-driver/latest/api/struct.CassCluster/#1a46b9816129aaa5ab61a1363489dccfd0
```
mikefero: Should be `http://docs.datastax.com/en/developer/cpp-driver/latest/api/struct.CassCluster/#function-cass_cluster_set_load_balance_dc_aware` in order to get to the `allow_remote_dcs_for_local_cl` parameter description

topics/metrics/README.md
```diff
@@ -33,8 +33,8 @@ host. This can occur when too many requests are in-flight for a single host.
 Connection timeouts occur when the process of establishing new connections is
 unresponsive (default: 5 seconds).
 
-[`cass_session_get_metrics()`]: http://datastax.github.io/cpp-driver/api/struct.CassSession/#1ab3773670c98c00290bad48a6df0f9eae
-[`CassMetrics`]: http://datastax.github.io/cpp-driver/api/struct.CassMetrics/
-[`requests`]: http://datastax.github.io/cpp-driver/api/struct.CassMetrics/#attribute-requests
-[`stats`]: http://datastax.github.io/cpp-driver/api/struct.CassMetrics/#attribute-stats
-[`errors`]: http://datastax.github.io/cpp-driver/api/struct.CassMetrics/#attribute-errors
+[`cass_session_get_metrics()`]: http://docs.datastax.com/en/developer/cpp-driver/latest/api/struct.CassSession/#1ab3773670c98c00290bad48a6df0f9eae
```
mikefero: Should be `http://docs.datastax.com/en/developer/cpp-driver/latest/api/struct.CassSession/#function-cass_session_get_metrics`

mpenick:
Landed in 1b9f98ee3cf7a9ffab38e3e21acaa47c4ce39b9f

PR #405, opened 2018-02-23
Port Boost test conversions to Google test framework

mikefero:


mikefero:
NOTE: The failing TravisCI tests are for libuv v0.10.x. These can and should be ignored for now due to [CPP-515](https://datastax-oss.atlassian.net/browse/CPP-515).

PR #406, opened 2018-03-22
CPP-578 - Add support for `NO_COMPACT` startup option

mikefero:
Starting with Apache Cassandra v4.0, Thrift and COMPACT STORAGE is no
longer supported. For uninterrupted cluster upgrades, the C/C++ driver
supports the 'NO_COMPACT' startup option. Enabling this feature using
cass_cluster_set_no_compact() will have same effect as
'DROP COMPACT STORAGE', but only for the connected session.


mikefero:
This PR is based off of #405 and has the same issue with libuv 0.10.x; the TravisCI failures can be ignored for now due to [CPP-515](https://datastax-oss.atlassian.net/browse/CPP-515).

PR #413, opened 2018-05-05
CPP-103 Compression support

vhavel:


src/request_callback.cpp
```diff
@@ -91,6 +99,23 @@ int32_t RequestCallback::encode(int version, int flags, BufferVec* bufs) {
   return length + header_size;
 }
 
+ICompressor::Buffer RequestCallback::compress(const BufferVec& bufs, ICompressor* compressor) const {
```
mikefero: I am not sure if this method is needed; operate off of the original `BufferVec` and allow for the new compressed `BufferVec` to be returned from the compression algorithm along with a flag to determine if the compression was successful.

src/compression.cpp
```diff
@@ -0,0 +1,136 @@
...
+  virtual const char* get_method_name() const {
+    return "snappy";
+  };
+};
+}
+
+#endif
+
+namespace cass {
+ICompressor::Ptr get_compressor(const std::list<std::string>& methods,
```
mikefero: To simplify this how about moving the initialization of the compression to the `Config` object.

src/connection.cpp
```diff
@@ -269,14 +269,16 @@ void Connection::connect() {
 }
 
 bool Connection::write(const RequestCallback::Ptr& callback, bool flush_immediately) {
-  int32_t result = internal_write(callback, flush_immediately);
+  int32_t result = internal_write(callback, flush_immediately, compressor_.get());
```
mikefero: A compression instance already exists in this class and should be called directly; extra `compression` parameter added to methods in this class and not needed and can be reverted.

mpenick:
Thanks again for the PR, but htere hasn't been any follow up on this so I'm going to close it for now. We can reopen it in another PR in the future.

PR #416, opened 2018-07-06
Fixing recovery of remote dc host

bvandersson:
Current implementation of dc_aware_policy removes host at on_down, destroying possibility to recover. The distance function always return CASS_HOST_DISTANCE_IGNORE.

### DIFF
--- a/src/dc_aware_policy.cpp
+++ b/src/dc_aware_policy.cpp
@@ -94,11 +94,21 @@ void DCAwarePolicy::on_remove(const Host::Ptr& host) {
 }
 
 void DCAwarePolicy::on_up(const Host::Ptr& host) {
-  on_add(host);
+  const std::string& dc = host->dc();
+  if (dc == local_dc_) {
+    change_host_status(local_dc_live_hosts_, host, Host::UP);
+  } else {
+    per_remote_dc_live_hosts_.change_host_status(dc, host, Host::UP);
+  }
 }
 
 void DCAwarePolicy::on_down(const Host::Ptr& host) {
-  on_remove(host);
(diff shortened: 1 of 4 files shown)


src/dc_aware_policy.cpp
```diff
@@ -94,11 +94,21 @@ void DCAwarePolicy::on_remove(const Host::Ptr& host) {
 }
 
 void DCAwarePolicy::on_up(const Host::Ptr& host) {
-  on_add(host);
+  const std::string& dc = host->dc();
```
mpenick: I'm pretty sure the driver shares the same instance of hosts throughout the driver and they should be marked UP/DOWN by the control connection and/or session. I think this could be fixed by just removing `on_add()` and `on_remove()` from these methods.

mpenick:
Thanks for the pull request. This is a very good catch. Were you able to sign the CLA?

mpenick:
Does this PR https://github.com/datastax/cpp-driver/pull/411 solve your bind issue or is this something different?

bvandersson:
This is exactly what we need, I will use it instead of our patch.


Thanks,

Björn

________________________________

mpenick:
Working on a fix here: https://github.com/datastax/cpp-driver/pull/427

mpenick:
We're adding a fix to the v2.10.0 release (https://github.com/datastax/cpp-driver/pull/427). This takes a different approach. The `distance()` function is not used to determine if a node should be added or marked as up.

mpenick:
Fixed as part of CPP-687. Please feel free to open another PR or a JIRA ticket if this doesn't fix your issue.

PR #420, opened 2018-08-09
master: CPP-669 - Result response's metadata can reference a prepared response that's already been freed

mpenick:
…ponse that's already been freed


mpenick:
Maybe there's a better way to do this? Should `ResultMetadata` just always carry a reference to the data (could be safer)?

mikefero:
This is being merged in from 1.6.0 merge into 2.10.0 [here](https://github.com/datastax/cpp-driver/commit/daebbfcbaa2b73f108c916760c812908abdecd5a). Closing this PR in favor of #421.

PR #424, opened 2018-09-28
CPP-674 - Fix memaccess in rapidjson

mikefero:
* This addresses the build issues with GCC 8


mpenick:
Does it make sense to just pull in the latest version? There's a fix on master: https://github.com/Tencent/rapidjson/commit/fa5963a2f5b231ee2babff771f169ccca22870ed

I'm pretty sure it also fixes our two other patches too. Maybe at the least we can create a JIRA to pull in the latest.

mikefero:
I would prefer we create a ticket and bring in the latest during another release cycle. Created [CPP-686](https://datastax-oss.atlassian.net/browse/CPP-686)

mpenick:
We should pull in that full patch then: https://github.com/Tencent/rapidjson/commit/fa5963a2f5b231ee2babff771f169ccca22870ed

PR #427, opened 2018-10-03
CPP-687 - Fix recovery of remote DC host

mpenick:
The DC-aware load balancing policy depends on an internal host map to determine host distance. The problem was the host distance was being checked to determine whether a node should be added or marked as up.  This fix always adds and marks hosts up even if they're ignored. The distance is only used to determine if a host should have a connection pool.

### DIFF
--- a/gtests/src/unit/mockssandra.cpp
+++ b/gtests/src/unit/mockssandra.cpp
@@ -1807,11 +1807,18 @@ cass::String SchemaChangeEvent::encode(SchemaChangeEvent::Target target, SchemaC
 
 void Cluster::init(AddressGenerator& generator,
                    ClientConnectionFactory& factory,
-                   size_t num_nodes) {
+                   size_t num_nodes_dc1, size_t num_nodes_dc2) {
   MT19937_64 token_rng;
-  for (size_t i = 0; i < num_nodes; ++i) {
+  for (size_t i = 0; i < num_nodes_dc1; ++i) {
     Address address(generator.next());
     Server server(Host(address, "dc1", "rack1", token_rng),
+                  internal::ServerConnection::Ptr(
+                    Memory::allocate<internal::ServerConnection>(address, factory)));
+    servers_.push_back(server);
+  }
+  for (size_t i = 0; i < num_nodes_dc2; ++i) {
+    Address address(generator.next());
+    Server server(Host(address, "dc2", "rack1", token_rng),
                   internal::ServerConnection::Ptr(
                     Memory::allocate<internal::ServerConnection>(address, factory)));
     servers_.push_back(server);
--- a/gtests/src/unit/mockssandra.cpp
+++ b/gtests/src/unit/mockssandra.cpp
@@ -1807,7 +1807,8 @@ cass::String SchemaChangeEvent::encode(SchemaChangeEvent::Target target, SchemaC
 
 void Cluster::init(AddressGenerator& generator,
                    ClientConnectionFactory& factory,
-                   size_t num_nodes_dc1, size_t num_nodes_dc2) {
+                   size_t num_nodes_dc1,
+                   size_t num_nodes_dc2) {
   MT19937_64 token_rng;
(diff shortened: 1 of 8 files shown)


mpenick:
This is only `wip` because I haven't fully tested or added a test. Feel free to access the fix.

gtests/src/unit/mockssandra.cpp
```diff
@@ -1807,15 +1807,22 @@ cass::String SchemaChangeEvent::encode(SchemaChangeEvent::Target target, SchemaC
 
 void Cluster::init(AddressGenerator& generator,
                    ClientConnectionFactory& factory,
-                   size_t num_nodes) {
+                   size_t num_nodes_dc1, size_t num_nodes_dc2) {
```
mikefero: nit: `size_t num_nodes_dc1, size_t num_nodes_dc2) {` =>

```
size_t num_nodes_dc1,
size_t num_nodes_dc2) {
```

gtests/src/unit/mockssandra.hpp
```diff
@@ -1119,7 +1119,7 @@ class Cluster {
 protected:
   void init(AddressGenerator& generator,
             ClientConnectionFactory& factory,
-            size_t num_nodes);
+            size_t num_nodes_dc1, size_t num_nodes_dc2);
```
mikefero: nit: same

gtests/src/unit/tests/test_cluster.cpp
```diff
@@ -668,3 +694,89 @@ TEST_F(ClusterUnitTest, InvalidSsl) {
...
+  connect_future->cluster()->notify_down(remote_address);
+
+  // Notify the remote host as up
+  connect_future->cluster()->notify_up(remote_address);
+
+  // Verify that the remote host was marked as up
+  ASSERT_TRUE(up_future->wait_for(WAIT_FOR_TIME));
+  EXPECT_EQ(remote_address, listener->address());
+
+  cluster.stop(1); // Stop local node to very that remote host is tried for reconnection.
```
mikefero: `very` => `verify`

gtests/src/unit/tests/test_cluster.cpp
```diff
@@ -668,3 +694,89 @@ TEST_F(ClusterUnitTest, InvalidSsl) {
...
+  ContactPointList contact_points;
+  contact_points.push_back("127.0.0.1");
+
+  Future::Ptr connect_future(Memory::allocate<Future>());
+  ClusterConnector::Ptr connector(Memory::allocate<ClusterConnector>(contact_points,
+                                                                     PROTOCOL_VERSION,
+                                                                     bind_callback(on_connection_connected, connect_future.get())));
+
+  ClusterSettings settings;
+  settings.load_balancing_policy.reset(Memory::allocate<DCAwarePolicy>("dc3", 0, false)); // Invalid DC and no using remote hosts
```
mikefero: I know `dc3` is not valid, but `dc2` is also invalid. What do you think about changing this to just `invalid_dc`?

`no using` => `not using`

src/cluster_connector.cpp
```diff
@@ -230,28 +231,26 @@ void ClusterConnector::on_connect(ControlConnector* connector) {
...
-    // Gather the available hosts for the load balancing policies
-    HostMap available_hosts;
-    for (HostMap::const_iterator it = hosts.begin(), end = hosts.end();
-         it != end; ++it) {
-      if (!is_host_ignored(policies, it->second)) {
-        available_hosts[it->first] = it->second;
+    ScopedPtr<QueryPlan> query_plan(default_policy->new_query_plan("", NULL, NULL));
+    if (!query_plan->compute_next()) { // No hosts in the query plan
+      const char* message;
+      if (dynamic_cast<DCAwarePolicy::DCAwareQueryPlan*>(query_plan.get()) != NULL) { // Check if DC-aware
```
mikefero: Nice ... thanks for taking care of that `TODO`

PR #428, opened 2018-10-08
CPP-692 - Fix C++98 compiler issues

mpenick:


gtests/src/unit/tests/test_copy_on_write.cpp
```diff
@@ -23,7 +23,7 @@
 #include "vector.hpp"
 
 TEST(CopyOnWriteUnitTest, Simple) {
-  cass::Vector<int>* ptr = cass::Memory::allocate<cass::Vector<int>>();
```
mikefero: This has always bothered me that older compilers treat this as a failure.

PR #433, opened 2018-10-15
CPP-688 - Abstract protocol handling

mpenick:


src/protocol.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+int ProtocolVersion::value() const {
+  return value_;
+}
+
+bool ProtocolVersion::is_valid() const {
+  if (value_ < CASS_LOWEST_SUPPORTED_PROTOCOL_VERSION) {
+    LOG_ERROR("Protocol version %s is lower than the lowest supported "
+              " protocol version %s",
+              to_string().c_str(),
```
mikefero: Infinite recursion could happen here since `to_string()` checks if `if(is_valid())`

src/protocol.cpp
```diff
@@ -66,7 +69,11 @@ bool ProtocolVersion::is_beta() const {
 String ProtocolVersion::to_string() const {
   if (value_ > 0) {
     OStringStream ss;
-    ss << "v" << value_;
+    if (value_ & DSE_PROTOCOL_VERSION_BIT) {
+      ss << "DSEv" << (value_ & DSE_PROTOCOL_VERSION_MASK);
```
mikefero: Good catch!

mpenick: Thanks. It looks better when logged.

PR #439, opened 2019-03-01
CPP-755 UDT schema metadata not properly populated/updated

mpenick:
A typo in `ControlConnection` prevents UDT schema metadata from being
properly updated.

### DIFF
--- a/examples/udt/udt.c
+++ b/examples/udt/udt.c
@@ -248,8 +248,6 @@ int main(int argc, char* argv[]) {
     return -1;
   }
 
-  schema_meta = cass_session_get_schema_meta(session);
-
   execute_query(session,
                 "CREATE KEYSPACE examples WITH replication = { \
                            'class': 'SimpleStrategy', 'replication_factor': '3' }");
@@ -262,6 +260,8 @@ int main(int argc, char* argv[]) {
 
   execute_query(session,
                 "CREATE TABLE examples.udt (id timeuuid, address frozen<address>, PRIMARY KEY(id))");
+
+  schema_meta = cass_session_get_schema_meta(session);
 
   insert_into_udt(session);
   select_from_udt(session);
--- a/src/control_connector.cpp
+++ b/src/control_connector.cpp
@@ -327,7 +327,7 @@ void ControlConnector::handle_query_schema(SchemaConnectorRequestCallback* callb
   schema_.views = callback->result("views");
   schema_.columns = callback->result("columns");
   schema_.indexes = callback->result("indexes");
-  schema_.user_types = callback->result("types");
+  schema_.user_types = callback->result("user_types");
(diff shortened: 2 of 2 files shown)


mpenick:
Verified the fix using the `udt.c` example, but this still needs a regression test.

chakri-stark:
The test with udt.c will fail if the keyspace already exists.
Moving cass_session_get_schema_meta inside insert call is required.

examples/udt/udt.c
```diff
@@ -29,6 +29,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <unistd.h>
```
mikefero: Needs to be updated for Windows; maybe something like this:

```diff
diff --git a/examples/udt/udt.c b/examples/udt/udt.c
index 21e63bde..3d9c9fed 100644
--- a/examples/udt/udt.c
+++ b/examples/udt/udt.c
@@ -29,7 +29,13 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
+#ifndef _WIN32
+# include <unistd.h>
+# define SLEEP(x) sleep(x)
+#else
+# include <windows.h>
+# define SLEEP(x) Sleep(x * 1000) // milliseconds
+#endif

 #include "cassandra.h"

@@ -247,7 +253,7 @@ void execute_schema_query(CassSession* session, const char* query) {

   version = schema_meta_version(session);
   while (version == existing_version && count < 20) {
-    sleep(1);
+    SLEEP(1);
     count++;
     version = schema_meta_version(session);
   }

```

examples/udt/udt.c
```diff
@@ -248,20 +274,20 @@ int main(int argc, char* argv[]) {
...
+                       "CREATE TYPE examples.address (street text, city text, zip int, phone set<frozen<phone_numbers>>)");
 
-  execute_query(session,
-                "CREATE TYPE examples.address (street text, city text, zip int, phone set<frozen<phone_numbers>>)");
+  execute_schema_query(session,
+                       "CREATE TABLE examples.udt (id timeuuid, address frozen<address>, PRIMARY KEY(id))");
 
-  execute_query(session,
-                "CREATE TABLE examples.udt (id timeuuid, address frozen<address>, PRIMARY KEY(id))");
+  schema_meta = cass_session_get_schema_meta(session);
```
chakri-stark: I hope adding this one line will update the schema meta, as it is not required to update for each schema change in this scenario.

examples/udt/udt.c
mpenick: Thanks, simplified my solution. I think my original fix was a bit too paranoid.

examples/udt/udt.c
mpenick: Yes, you're correct.

PR #440, opened 2019-04-10
Fix generation of shlib deps file for debian packages 

jirkasilhan:
Adds current version for correct transitive dependencies.

Without this patch shlib file doesnt contain minimal compatible version of cassandra-cpp-driver and transitive dependencies may be broken.

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -32,5 +32,8 @@ override_dh_strip:
 	sed -i s/@DEB_HOST_MULTIARCH@/$(DEB_HOST_MULTIARCH)/ debian/cassandra-cpp-driver-dev/usr/lib/$(DEB_HOST_MULTIARCH)/pkgconfig/*.pc
 	sed -i s/@DEB_VERSION_UPSTREAM@/$(DEB_VERSION_UPSTREAM)/ debian/cassandra-cpp-driver-dev/usr/lib/$(DEB_HOST_MULTIARCH)/pkgconfig/*.pc
 
+override_dh_makeshlibs:
+	dh_makeshlibs -V
+
 override_dh_auto_test:
 override_dh_auto_clean:


mpenick:
Thanks! I'll take a look today and get this merged in.

jirkasilhan:
Ok, i'm looking forward for the merge...

packaging/debian/rules
```diff
@@ -32,5 +32,8 @@ override_dh_strip:
 	sed -i s/@DEB_HOST_MULTIARCH@/$(DEB_HOST_MULTIARCH)/ debian/cassandra-cpp-driver-dev/usr/lib/$(DEB_HOST_MULTIARCH)/pkgconfig/*.pc
 	sed -i s/@DEB_VERSION_UPSTREAM@/$(DEB_VERSION_UPSTREAM)/ debian/cassandra-cpp-driver-dev/usr/lib/$(DEB_HOST_MULTIARCH)/pkgconfig/*.pc
 
+override_dh_makeshlibs:
+       dh_makeshlibs -V
```
mpenick: This should be a tab here instead of spaces for maximum compatibility. I get an error on my system.

```
 fakeroot debian/rules clean
debian/rules:36: *** missing separator.  Stop.
dpkg-buildpackage: error: fakeroot debian/rules clean subprocess returned exit status 2
debuild: fatal error at line 1152:
```

jirkasilhan: This patch is not fix of driver dependencies. It fixes the generation of shlibs file distributed in driver package.

Default behavior of makeshlibs depends on convention of placing major version of library in binary package name - i.e.- for development package mylib-dev with version 1.2.3 you have binary package named mylib1. 

This convention solves the problem of linking binary incompatible versions of library to your executable, so you can make dependency on mylib based just on the package name (mylib1). This is default behaviour of dh_makeshlibs - it will generate file of dependencies and name of binary package.

But this is different in cassandra cpp driver - you have binary incompatible versions with same package name (but different version). In this case, you need to set the version of driver in shlibs file - so dependent packages will depend on version which was used during their build.

I will fix the tabs/spaces...

mpenick: Thanks for the explanation!

jirkasilhan: Ok, whitespace is fixed...

PR #442, opened 2019-04-28
Fixed the location of the '[in]' information for Doxygen.

AlexisWilke:
Somehow this one entry had the `[in]` after the variable names instead of just after `\param`.


mpenick:
Thanks!

PR #443, opened 2019-05-22
Changed CMAKE_SOURCE_DIR to CMAKE_CURRENT_SOURCE_DIR

accelerated:
**Description**
Changing `CMAKE_SOURCE_DIR ` to `CMAKE_CURRENT_SOURCE_DIR` allows the cmake project to be called as a subproject by another parent CMakeLists.txt. If `CMAKE_SOURCE_DIR ` is used, it is _global_ per cmake project and causes all file paths to be wrongly set.


mpenick:
Thanks! :+1:

accelerated:
That was fast!

PR #444, opened 2019-05-23
Added header subdir install capability

accelerated:
**Description**
Add two configuration options for installing `cassandra.h` header:
* `CASS_INSTALL_HEADER_IN_SUBDIR=ON/OFF (default OFF)` : This installs the header under `include/cassandra` instead of `include`. This option is needed since not all systems prefer to have a flat include directory. The current CMakeLists.txt makes this assumption.
* `CASS_INSTALL_HEADER_SUBDIR_NAME=<name>` : If `CASS_INSTALL_HEADER_IN_SUBDIR=ON` then it overrides the subdirectory name. Header file will be installed under `include/<name>`. Could be for instance `include/cassandra-cppdriver`.

**NOTE**: Both these options are enabled only if `CASS_INSTALL_HEADER=ON`


mpenick:
Thanks! :+1:

PR #445, opened 2019-05-30
Changed pkg_config templates to use the library variable name

accelerated:
**Description**
* Changed pkg_config templates to use the library variable name instead of being hard-coded


mikefero:
Nice and clean, thanks for the contribution.

PR #446, opened 2019-05-31
Sort of explicit_bzero proposal.

devnexen:
To provide a non optimisable version, compiler-wise, of memset
for security purposes. For now limiting the scope to md5 class.

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -11224,6 +11224,17 @@ cass_alloc_set_functions(CassMallocFunction malloc_func,
                          CassReallocFunction realloc_func,
                          CassFreeFunction free_func);
 
+/**
+ * Clean dst buffer to len size
+ *
+ * Similar as bzero but not optimisable regardless
+ * of the compiler optimisation level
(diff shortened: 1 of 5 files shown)


mpenick:
Thanks for the PR, but we're not using `Md5` as hash for anything that's security related so this is not a big concern for us. The two places we use `Md5` are for generating random-ish data for uuids and for token aware routing (but it's almost never used for this case). If we did require a hash for security purposes we would choose something other than MD5.

include/cassandra.h
```diff
@@ -11224,6 +11224,17 @@ cass_alloc_set_functions(CassMallocFunction malloc_func,
...
+/**
+ * Clean dst buffer to len size
+ *
+ * Similar as bzero but not optimisable regardless
+ * of the compiler optimisation level
+ * @param[in] dst
+ * @param[in] len
+ */
+CASS_EXPORT void
+cass_secure_zero(void * dst,
```
mpenick: In general, the API tends not to expose generic functionality externally.

PR #447, opened 2019-06-04
CPP-785 - Concurrent execution async example

mikefero:


examples/concurrent_executions/concurrent_executions.c
```diff
@@ -0,0 +1,193 @@
...
+*/
+
+#include <stdio.h>
+#include <string.h>
+
+#include "cassandra.h"
+
+#include <uv.h>
+
+#define CONCURRENCY_LEVEL 32 /* Maximum amount of parallel async executions (threads) */
```
mpenick: This should probably be `NUM_THREADS` and should be much smaller than 32 or we could just eliminate threads entirely from this example to simplify it to focus on doing concurrent asynchronous requests. I think that `CONCURRENCY_LEVEL` is a better name for outstanding asynchronous requests.

examples/concurrent_executions/concurrent_executions.c
```diff
@@ -0,0 +1,193 @@
...
+#define NUM_REQUESTS 10000
+
+typedef struct Concurrent_ {
```
mpenick: Not needed, use globals.

examples/concurrent_executions/concurrent_executions.c
mpenick: Maybe a 3rd way is using synchronous requests from multiple threads.

mpenick:
It looks like https://github.com/datastax/nodejs-driver/blob/master/examples/concurrent-executions/execute-in-loop.js is doing what I call "chunking".

examples/concurrent_executions/concurrent_executions.c
mikefero: I chose the chunking method because I wanted to keep things off the callback for our examples as a "best practice" approach.

examples/concurrent_executions/concurrent_executions.c
```diff
@@ -0,0 +1,178 @@
...
+  CassFuture* futures[CONCURRENCY_LEVEL];
+  int num_requests = NUM_REQUESTS;
+
+  while (num_requests > 0) {
+    int i;
+    int num_outstanding_requests = CONCURRENCY_LEVEL;
+    if (num_requests < num_outstanding_requests) {
+      num_outstanding_requests = num_requests;
+    }
+
```
mpenick: Might be clearer if `num_requests` was updated here instead of below in the future check loop.

```c
num_request -= num_outstanding_requests;
```

PR #449, opened 2019-06-14
Added helpers for CassConsistency level

accelerated:
### Description
Added two helper functions for determining the next weaker consistency level and also for comparing two consistency levels. This is very useful when the application needs logic to retry a statement at a lower consistency based on previous failure(s).

New functions added:
1.  `cass_consistency_level_compare()`
2. `cass_next_weaker_consistency_level()`

New enum added:
1. `CassOperationType`

**Note**: This is a preliminary PR as i'm not 100% sure where the function definitions belong to. I put them under `utils.cpp` but they could go in a different location. Also there are no tests yet since I wasn't sure if the PR was going to approved to begin with. I can add some test cases inside `test_consistency.cpp` if need be.
A much simpler way to deal with this would be to re-arrange the `CassConsistency` enums in increasing order, but I feel this would break the API and there might be implementations out there relying on the current hard-coded values, hence these two helper methods.

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -913,18 +913,22 @@ typedef void(*CassHostListenerCallback)(CassHostListenerEvent event,
  * Gets the next weaker consistency level. Calling this function with the weakest
  * level will return the same value.
  *
+ * @param[in] dc_replication_factor The replication factor used per data center.
  * @param[in] op Indicates if it's a read or write operation.
  * @param[in] stronger The consistency level.
  * @return The next weaker consistency level. If there is a mismatch between the operation
  *         type and the consistency level provided, CASS_CONSISTENCY_UNKNOWN is returned.
  */
 CASS_EXPORT CassConsistency
-cass_next_weaker_consistency_level(CassOperationType op, CassConsistency stronger);
+cass_next_weaker_consistency_level(size_t dc_replication_factor,
+                                   CassOperationType op,
+                                   CassConsistency stronger);
 
 
 /**
  * Compares two consistency levels.
  *
+ * @param[in] dc_replication_factor The replication factor used per data center.
  * @param[in] op Indicates if it's a read or write operation.
  * @param[in] stronger The stronger consistency level.
  * @param[in] weaker The weaker consistency level.
@@ -932,7 +936,10 @@ cass_next_weaker_consistency_level(CassOperationType op, CassConsistency stronge
  *         if one of them is CASS_CONSISTENCY_UNKNOWN, this function returns cass_false.
  */
 CASS_EXPORT cass_bool_t
-cass_consistency_level_compare(CassOperationType op, CassConsistency stronger, CassConsistency weaker);
+cass_consistency_level_compare(size_t dc_replication_factor,
+                               CassOperationType op,
+                               CassConsistency stronger,
+                               CassConsistency weaker);
 
 /***********************************************************************************
  *
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -718,6 +718,11 @@ typedef enum CassError_ {
   /* @endcond*/
 } CassError;
 
+typedef enum CassOperationType_ {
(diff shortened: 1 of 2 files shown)


mpenick:
Thanks of for the PR, but the ordering of consistency levels is application specific and can't be generalized in this way. For example, 
 `CASS_CONSISTENCY_LOCAL_QUORUM` --> `CASS_CONSISTENCY_THREE` doesn't make sense if the replication factor is less than five and that's pretty uncommon for a data model.

mpenick:
> This is very useful when the application needs logic to retry a statement at a lower consistency based on previous failure(s).

Consistency level along with replication factor should be chosen based on the availability and consistency guarantees of your application and shouldn't be lowered based on node outages. For example, if your application's can tolerate the lower consistency of `CASS_CONSISTENCY_ONE` instead of  `CASS_CONSITENCY_QUORUM` then use `CASS_CONSISTENCY_ONE`.

accelerated:
> Thanks of for the PR, but the ordering of consistency levels is application specific and can't be generalized in this way. For example,
> `CASS_CONSISTENCY_LOCAL_QUORUM` --> `CASS_CONSISTENCY_THREE` doesn't make sense if the replication factor is less than six and that's pretty uncommon for a data model.

I agree with this, but just in the same way that the application can determine if the consistency should be lowered to say CASS_CONSISTENCY_TWO or ONE (because the replication factor is less than 6), it can also skip the THREE case if it knows it does not apply. These helpers are not generic across the board, they simply help establish, should the case apply, a guideline for lowering the consistency. An application can always make an improper determination (even today) and run a query at a consistency level which would "make no sense" - such as THREE in your example. That doesn't mean that the API should not allow overriding the consistency on a per-statement basis! 

>Consistency level along with replication factor should be chosen based on the availability and consistency guarantees of your application and shouldn't be lowered based on node outages.

I understand but this is purely application dependent and can be no more than a guideline. Based on the tests we have conducted and running in production with various outages, some would prefer a different approach, such as gradually lowering the consistency.

accelerated:
Perhaps a sensible thing would be to also pass-in the replication factor used, therefore consistencies which don't apply will be skipped.

accelerated:
Please take a look. The consistency levels should now be properly reported.

mpenick:
> These helpers are not generic across the board, they simply help establish, should the case apply, a guideline for lowering the consistency.

I'd prefer not to include this as part of the cpp-driver's public API. As a general-purpose Cassandra/DSE driver we try to provide generic mechanisms for our users and avoid providing policies. The ordering outlined in `cass_next_weaker_consistency_level()` may work for your specific application, but it will likely not work well for other applications and is counter to our best practice of choosing the lowest consistency level for the specific guarantees of an application's availability/consistency.

Other issue more specific issues:
* It's not clear if `dc_replication_factor` is the local DC or the remote DC(s). From the header docs it seems like it's the replication factor of each DC in the cluster, but not all applications use the same replication factor in all DCs.
* Going from `EACH_QUORUM` --> `QUORUM` seems pretty dangerous because `EACH_QUORUM` give you a guarantees about data replicas in the remote DC(s) and with `QUORUM` all the replicas could happen in the local DC only. 
* Going from `LOCAL_ONE` --> `SERIAL`  is an increase in the consistency level, but the bigger issue is `SERIAL` is sort of tangential to other read consistency level because it implies the use of lightweight transactions.

accelerated:
Thanks @mpenick for the explanations. My downgrade policy was based on the following [document](https://docs.datastax.com/en/archived/cassandra/3.0/cassandra/dml/dmlConfigConsistency.html#dmlConfigConsistency__dml-config-read-consistency) which states that all the levels there are ordered from stronger to weaker.  In any case, based on your comment, are you saying that `EACH_QUORUM` is NOT a stronger consistency than `QUORUM`? I'm ok if these are not included in your lib...I'll keep them privately in our app (with changes based on your recommendations), although I would have preferred something more official.

PR #451, opened 2019-07-16
fixed link in README.md

beccam:
Added "/" to the end of download server address link such that it would work.


mpenick:
Thanks!

PR #452, opened 2019-08-02
Allow random to work with 0.

joeyhub:
This doesn't solve the problem of somehow you allow it to initiate with no contact points rather than noop or graceful error (with php-driver try to connect with host as empty string).

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -53,6 +53,11 @@ Random::~Random() { uv_mutex_destroy(&mutex_); }
 
 uint64_t Random::next(uint64_t max) {
   ScopedMutex l(&mutex_);
+
+  if (max == 0) {
+    return 0;
+  }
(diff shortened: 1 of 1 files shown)


joeyhub:
```
  const uint64_t limit = CASS_UINT64_MAX - CASS_UINT64_MAX % max;
  uint64_t r;
  do {
    r = rng_();
  } while (r >= limit);
  return r % max;
```
Return 0 might be wrong. Instead...

```
  const uint64_t limit = max ? CASS_UINT64_MAX - CASS_UINT64_MAX % max : CASS_UINT64_MAX;
  return max ? r % max : r;
```

mpenick:
Created a story to fix the error handling around no contact points: https://datastax-oss.atlassian.net/browse/CPP-806

PR #453, opened 2019-08-16
CPP-754: remove user-defined copy contructors

eevans:
Under GCC9 `-Werror=deprecated-copy` makes an implicit declaration of
a copy constructor an error if a user-provided one exists.

See also: https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/C_002b_002b-Dialect-Options.html#index-Wdeprecated-copy


mpenick:
Thanks @eevans .

remicollet:
Thanks.

I confirm this fixes build with GCC 9

PR #454, opened 2019-08-20
Add openssl to the required library list in pkg_config file

accelerated:
Add `openssl` to the required library list in pkg_config file when building a static library. Otherwise pkg_config complains if `CASS_USE_OPENSSL` is enabled. Generated .pc file is:

```
prefix=/usr/local
exec_prefix=/usr/local
libdir=/usr/local/lib64
includedir=/usr/local/include/cassandra-datastax

Name: cassandra
Description: A C/C++ client driver for Apache Cassandra
Version: 2.13.0
Requires: libuv openssl
Libs: -L${libdir} -lcassandra_static -lstdc++
Cflags:
URL: https://github.com/datastax/cpp-driver/
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,12 @@ if(CASS_BUILD_STATIC)
   endif()
   CassConfigureStatic("CASS")
 
+  # Set pkg_config required libraries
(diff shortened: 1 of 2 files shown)


mikefero:
Thanks for the contribution.

PR #455, opened 2019-08-24
Tampering timing attack with CRYPTO_memcmp usage.

devnexen:
even though the name is a bit misleading, closer to bcmp in fact,
also the main difference, it does not stop going through the buffer
upon the first mismatch.


devnexen:
Yes I ve been pushing too far for no gain. :-)

PR #461, opened 2019-11-02
CPP-852 Remove std::ptr_fun<T> and std::not1

fredrikslattman:
std::ptr_fun<T> has been deprecated since
C++11, and will be removed in C++17. Compiling
with Clang 10+ will yeild a new error which
is being treated as an error.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -104,10 +104,10 @@ String implode(const Vector<String>& vec, const char delimiter /* = ' ' */) {
 String& trim(String& str) {
   // Trim front
   str.erase(str.begin(),
-            std::find_if(str.begin(), str.end(), std::not1(std::ptr_fun<int, int>(::isspace))));
+            std::find_if(str.begin(), str.end(), [](unsigned char c) { return !::isspace(c); }));
   // Trim back
   str.erase(
-      std::find_if(str.rbegin(), str.rend(), std::not1(std::ptr_fun<int, int>(::isspace))).base(),
+      std::find_if(str.rbegin(), str.rend(), [](unsigned char c) { return !::isspace(c); }).base(),
       str.end());
   return str;
 }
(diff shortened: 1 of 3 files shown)


src/utils.cpp
```diff
@@ -104,10 +104,10 @@ String implode(const Vector<String>& vec, const char delimiter /* = ' ' */) {
 String& trim(String& str) {
   // Trim front
   str.erase(str.begin(),
-            std::find_if(str.begin(), str.end(), std::not1(std::ptr_fun<int, int>(::isspace))));
+            std::find_if(str.begin(), str.end(), [](unsigned char c) { return !::isspace(c); }));
```
mikefero: Lambdas are a C++11 feature. We currently support older compilers that do not support C++11 features.

fredrikslattman: I understand. I don't have a C++03 compiler at hand, so I'll close the PR. Thanks for the review either way! :)

mpenick:
I think this would solve all the things.

```cpp
diff --git a/cpp-driver/src/utils.cpp b/cpp-driver/src/utils.cpp
index 2e105c30..208b9091 100644
--- a/cpp-driver/src/utils.cpp
+++ b/cpp-driver/src/utils.cpp
@@ -101,14 +101,13 @@ String implode(const Vector<String>& vec, const char delimiter /* = ' ' */) {
   return str;
 }
 
+bool not_isspace(int c) { return !::isspace(c); }
+
 String& trim(String& str) {
   // Trim front
-  str.erase(str.begin(),
-            std::find_if(str.begin(), str.end(), std::not1(std::ptr_fun<int, int>(::isspace))));
+  str.erase(str.begin(), std::find_if(str.begin(), str.end(), not_isspace));
   // Trim back
-  str.erase(
-      std::find_if(str.rbegin(), str.rend(), std::not1(std::ptr_fun<int, int>(::isspace))).base(),
-      str.end());
+  str.erase(std::find_if(str.rbegin(), str.rend(), not_isspace).base(), str.end());
   return str;
 }
```

fredrikslattman:
Got some time over, but I killed my previous fork, so here is the new PR: https://github.com/datastax/cpp-driver/pull/462

PR #462, opened 2019-11-14
CPP-852 Remove std::ptr_fun<T> and std::not1

fredrikslattman:
std::ptr_fun<T> has been deprecated since C++11 [1], and std::not1
has been deprecated since C++17 [2]. Both can be removed without
loosing any functionality.

When using the driver through a package manager you might need to
compile it from source, and this enables the source to be compiled
with -std=c++17.

[1] https://en.cppreference.com/w/cpp/utility/functional/ptr_fun
[2] https://en.cppreference.com/w/cpp/utility/functional/not1


mpenick:
We really appreciate the PR, but this change has already landed in https://github.com/datastax/cpp-driver/commit/3eeb974f50611e0263db07f7acbc97b1b96e0da0.

PR #463, opened 2019-11-25
CPP-879 Allow remote hosts to come back up

kw217:
When a host comes back up after being down, a request processor (unlike a control connection) only notifies the load balancing policy if the host is not ignored.

Sadly some policies ignore remote hosts that are down.

In combination, this means that remote hosts can never be used by a request (processor) after having once gone down.

This fix resolves this, by always notifying a policy of all host-up events, regardless of the distance of that host.

I've tested this locally, and it has the desired effect. My test case is:
* Start nodes in two datacenters. 
* Set up a token-aware, dc-aware policy.
* Take down the local DC. Observe traffic still goes to the remote DC.
* Take down the remote DC. Observe traffic fails (as expected).
* Bring up the remote DC. Expect traffic will start going to the remote DC.

In the last step, in 2.14.0, instead I observe that traffic continues to fail. Logs and netstat reveal the control connection is up to the remote DC, but it is not used by the request processor. This PR fixes this, so that traffic does go to the remote DC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+Unreleased
+===========
+
+Bug Fixes
+--------
+* Allow remote hosts to come back up even if policy ignores down hosts
+
 2.14.0
 ===========
 
--- a/src/request_processor.cpp
+++ b/src/request_processor.cpp
@@ -462,9 +462,7 @@ void RequestProcessor::internal_host_ready(const Host::Ptr& host) {
     LoadBalancingPolicy::Vec policies = load_balancing_policies();
     for (LoadBalancingPolicy::Vec::const_iterator it = policies.begin(); it != policies.end();
          ++it) {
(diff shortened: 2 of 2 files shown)


rwincewicz:
Nice, simple fix. LGTM!

mpenick:
Thanks for the PR. Sorry for the delay, I've been out. I'll work on investigating this issue today.

accelerated:
Hi @mpenick, could we have a minor release with this PR. Seems like an important one.

mpenick:
Change looks great. Thanks again.

mpenick:
@accelerated yes, highly likely. You're right. This is minor release worthy.

mpenick:
@kw217 @accelerated This fix is in recently posted release [v2.14.1](https://github.com/datastax/cpp-driver/releases/tag/2.14.1).

PR #465, opened 2020-01-15
fix cflags when build with CASS_INSTALL_HEADER_IN_SUBDIR on

remicollet:
With this patch

```
$ pkg-config cassandra -cflags
-I/usr/include/cassandra 
```


mpenick:
Hi. Good change! Could you add this to https://github.com/remicollet/cpp-driver/blob/issue-fixpc/packaging/cassandra_static.pc.in too? If not no big deal, I can pull this in and I'll add it.

Thanks for the PR.

remicollet:
> Could you add this ...

Done and squashed

remicollet:
Sadly... CASS_INSTALL_HEADER_IN_SUBDIR is broken in 2.15.1

PR #466, opened 2020-02-10
CPP-747 fixing inconsistency in USE statements

derek9gag:
The `USE <keyspace>` statements are inconsistently quoted in different places. It is impossible to use a quoted keyspace (for example keyspace with upper-case characters, see [CPP-747](https://datastax-oss.atlassian.net/browse/CPP-747)). 

This PR fixes a regression from v2.9.0. Keyspaces with uppercases likely never worked since v2.10.0 (9292b536d58c2ce1904afc72fa8be8acd8b9c927). 

This also resolves [PHP-224](https://datastax-oss.atlassian.net/browse/PHP-224).

See these lines:

 - https://github.com/datastax/cpp-driver/blob/bbbbd7bc3eaba1b10ad8ac6f53c41fa93ee718db/src/prepare_host_handler.cpp#L183
 - https://github.com/datastax/cpp-driver/blob/bbbbd7bc3eaba1b10ad8ac6f53c41fa93ee718db/src/pooled_connection.cpp#L40
 - https://github.com/datastax/cpp-driver/blob/bbbbd7bc3eaba1b10ad8ac6f53c41fa93ee718db/src/connector.cpp#L265

### DIFF
--- a/src/connector.cpp
+++ b/src/connector.cpp
@@ -262,7 +262,7 @@ void Connector::on_ready_or_set_keyspace() {
     finish();
   } else {
     connection_->write_and_flush(RequestCallback::Ptr(
-        new StartupCallback(this, Request::ConstPtr(new QueryRequest("USE " + keyspace_)))));
+        new StartupCallback(this, Request::ConstPtr(new QueryRequest("USE \"" + keyspace_ + "\"")))));
   }
 }
 
--- a/src/prepare_host_handler.cpp
+++ b/src/prepare_host_handler.cpp
@@ -180,7 +180,7 @@ void PrepareHostHandler::PrepareCallback::on_internal_timeout() {
 
 PrepareHostHandler::SetKeyspaceCallback::SetKeyspaceCallback(const String& keyspace,
                                                              const PrepareHostHandler::Ptr& handler)
-    : SimpleRequestCallback(Request::ConstPtr(new QueryRequest("USE " + keyspace)))
+    : SimpleRequestCallback(Request::ConstPtr(new QueryRequest("USE \"" + keyspace + "\"")))
     , handler_(handler) {}
 
 void PrepareHostHandler::SetKeyspaceCallback::on_internal_set(ResponseMessage* response) {
--- a/tests/src/unit/mockssandra.cpp
+++ b/tests/src/unit/mockssandra.cpp
@@ -1810,7 +1810,7 @@ void UseKeyspace::on_run(Request* request) const {
     query.erase(0, query.find_first_not_of(" \t"));
     if (query.substr(0, 3) == "USE" || query.substr(0, 3) == "use") {
       query.erase(0, 3);
-      query.erase(0, query.find_first_not_of(" \t"));
+      query.erase(0, query.find_first_not_of(" \t\""));
       if (query.substr(0, keyspace.size()) == keyspace) {
         String body;
         encode_int32(RESULT_SET_KEYSPACE, &body);


src/connector.cpp
```diff
@@ -262,7 +262,7 @@ void Connector::on_ready_or_set_keyspace() {
     finish();
   } else {
     connection_->write_and_flush(RequestCallback::Ptr(
-        new StartupCallback(this, Request::ConstPtr(new QueryRequest("USE " + keyspace_)))));
+        new StartupCallback(this, Request::ConstPtr(new QueryRequest("USE \"" + keyspace_ + "\"")))));
```
mpenick: I think this will break using case intensive keyspaces from the C API.

There are two cases:
1) User provided keyspaces from the C API these take the form `caseinsensitive` or `\"CaseSensitive\"`
2) Cassandra provided keyspaces (from its metadata) which are will be sent as having the correct case, but i think won't be wrap in the double quotes when case senstivie.  I need to verify this.

So it's possible that we should use the form `"USE " + keyspace`, but properly wrap Cassandra provided keyspaces in double quotes when needed?

mpenick: The other possibility is that we just need to use `"USE \"" + keyspace + "\"" in the right spots e.g. `PrepareHostHandler::SetKeyspaceCallback`, but not the other two spots as long as we can prove that its keyspace values never come from the C API. We'd also have to prove that the other two spot don't take values from Cassandra provided keyspaces.

derek9gag: I have tried those using `cqlsh`, assuming we have a `"CaseSensitive"` and a `case_insensitive`  keyspaces:

- `USE "CaseSensitive";` works
- `USE CaseSensitive;` fails
- `USE "case_insensitive";` works
- `USE case_insensitive;` works
- `USE CASE_INSENSITIVE;` works
- `USE "CASE_INSENSITIVE";` **fails**

Cassandra provided keyspaces such as `system` and `system_auth` are the same as an user case_insensitive keyspace.

Always quoting the namespace only fails when we specify the wrong casing (e.g. `USE "System";`).

When we retrieve the metadata, the keyspaces will be have the correct casing like `system` and `system_auth`, and adding a quote around will work.

derek9gag: I see there is one case-sensitive table (`system."IndexInfo"`), but I have not seen and am not aware of any case-sensitive keyspaces.

derek9gag: I tried listing the keyspace names from CassSession schema. The keyspace names are not quoted. I did that via the PHP wrapper:

```
Listing keyspace names in session->schema():
 UpperCase
 lowercase
 system
 system_auth
 system_distributed
 system_schema
 system_traces
```

mpenick: The keyspace can't be double quoted in `connector.cpp` (https://github.com/datastax/cpp-driver/pull/466/files#diff-0fc019678d7d3eecf2cb341a50df9945R265) because it breaks a few cases when you connect with a keyspace. 

That's what I was attempting to explain above.

```c
  { // Works
    CassError rc;
    CassSession* session = cass_session_new();
    CassFuture* future = cass_session_connect_keyspace(session, cluster,  "case_insensitive");
    rc = cass_future_error_code(future);
    assert(rc == CASS_OK && "Same case for insensitive should work");
    cass_future_free(future);
    cass_session_free(session);
  }

  { // Should work, but doesn't work: Received error response 'Keyspace 'Case_Insensitive' does not exist' (0x02002200)
    CassError rc;
    CassSession* session = cass_session_new();
    CassFuture* future = cass_session_connect_keyspace(session, cluster,  "Case_Insensitive");
    rc = cass_future_error_code(future);
    assert(rc == CASS_OK && "Different case for insensitive should work");
    cass_future_free(future);
    cass_session_free(session);
  }

  { // Fails (because it's double, double quoted): 'line 1:4 no viable alternative at input '""' (USE [""]...)' (0x02002000)
    CassError rc;
    CassSession* session = cass_session_new();
    CassFuture* future = cass_session_connect_keyspace(session, cluster, "\"CaseSensitive\"");
    rc = cass_future_error_code(future);
    assert(rc == CASS_OK && "Same case with quotes for sensitive should work");
    cass_future_free(future);
    cass_session_free(session);
  }

  { // Works, but shouldn't (it should require double quotes to work)
    CassError rc;
    CassSession* session = cass_session_new();
    CassFuture* future = cass_session_connect_keyspace(session, cluster, "CaseSensitive");
    rc = cass_future_error_code(future);
    assert(rc != CASS_OK && "Same case without quotes for sensitive should not work");
    cass_future_free(future);
    cass_session_free(session);
  }
```

mpenick: I have a potential fix on a branch here: https://github.com/datastax/cpp-driver/commit/72ef57d98fc890747731f7e1e961c765eb3ff75b

derek9gag: I see. I checked out your branch and it works fine.

mpenick: Added a follow up PR (w/ additional tests) https://github.com/datastax/cpp-driver/pull/467.

derek9gag: Thank you!

derek9gag:
See #467

PR #468, opened 2020-02-17
Fix typo in RequestProcessorInitializer::internal_initialize

m8mble:
Three `i`s just are not good enough to initialize properly ;)


mpenick:
Thanks!

PR #472, opened 2020-03-18
fix CASS_INSTALL_HEADER_IN_SUBDIR usage

remicollet:
Since 90a61971299259a8d1e8de7311936b28606ea56b#diff-af3b638bc2a3e6c650974192a53c7291  `PROJECT_NAME_STRING` is no more defined


remicollet:
Using both `CASS_INSTALL_HEADER_IN_SUBDIR` **and** `CASS_INSTALL_HEADER_SUBDIR_NAME` is a workaround to this issue

mikefero:
@remicollet Thanks for the fix. We will be cutting the 2.15.1 release with this fix in it.

PR #473, opened 2020-03-26
Add comparison operators for CassUuid

lgeorget:
Hello Cassandra developers

I've needed operator<() and operator==() implementations for CassUuid variables for a project in which I use CassUuid in std::set or as keys in std::map objects.
Of course, they're pretty trivial to implement but maybe it would be nice to have them in cassandra.h (even if they are for C++ only)?


mpenick:
Thanks for the operators. I'd prefer these live in a C++ specific header and could you add some tests?

lgeorget:
Hello (a few months later...)!

So, I've put the operators in a separate header and added tests. However, I've maintained the inclusion of the operators in the main include file (cassandra.h) in order to avoid having to include another header for the operators only. I'm not sure this if this is ok or not. Should we just leave C++-only code in the other header?

include/cassandra.h
```diff
@@ -144,6 +144,11 @@ typedef struct CassUuid_ {
   cass_uint64_t clock_seq_and_node;
 } CassUuid;
 
+#ifdef __cplusplus
```
mpenick, line 147: Please remove this block. This is broken without adding it to build and packaging.

lgeorget:
Hmm, yes... I've renamed the include file, added it to CMakeLists.txt and to the Debian packaging files list. I've tested the packaging under Debian but I'm not able to testHomebrew right now.

PR #474, opened 2020-03-26
Add cass_ssl_set_default_verify_paths API

m8mble:
Forwards SSL-configuration to use system default directories for finding certificate authorities.

Similar functionality is often desired / provided by similar SSL-context forwarding APIs, e.g. [boost ssl context][boost].

[boost]: https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/reference/ssl__context.html

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -4679,6 +4679,18 @@ cass_ssl_set_private_key_n(CassSsl* ssl,
                            const char* password,
                            size_t password_length);
 
+/**
+ * Configures the context to use the default directories
+ * for finding certification authority certificates.
+ *
+ * @public @memberof CassSsl
+ *
+ * @param[in] ssl
+ * @return CASS_OK if successful, otherwise an error occurred
+ */
+CASS_EXPORT CassError
+cass_ssl_set_default_verify_paths(CassSsl* ssl);
+
 /***********************************************************************************
  *
  * Authenticator
--- a/src/ssl.cpp
+++ b/src/ssl.cpp
@@ -65,6 +65,10 @@ CassError cass_ssl_set_private_key_n(CassSsl* ssl, const char* key, size_t key_l
   return ssl->set_private_key(key, key_length, password, password_length);
 }
 
+CassError cass_ssl_set_default_verify_paths(CassSsl* ssl) {
+  return ssl->set_default_verify_paths();
+}
+
 } // extern "C"
 
 template <class T>
--- a/src/ssl.hpp
+++ b/src/ssl.hpp
@@ -87,6 +87,7 @@ public:
   virtual CassError set_cert(const char* cert, size_t cert_length) = 0;
   virtual CassError set_private_key(const char* key, size_t key_length, const char* password,
                                     size_t password_length) = 0;
+  virtual CassError set_default_verify_paths() = 0;
 
 protected:
   int verify_flags_;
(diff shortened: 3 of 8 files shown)


mpenick:
Sorry for the delay. Thanks for the PR. Looks great, very thorough. 

Calling `SSL_CTX_set_default_verify_paths()` as default internally would be a bad idea because it might add CAs the application does not intend? 

Would you be able to add a unit test for this? (Looks like it might be possible to control the default paths with env. variables  e.g. `SSL_CERT_DIR`)

m8mble:
Thanks for your feedback.

I don't think `SSL_CTX_set_default_verify_paths` by default would be a good idea: For one, it would simply be a breaking change. And for two, it's probably preferable to stay as close as possible to the actual openssl API. 

I've added a basic unit test to demonstrate the intended behavior. Thanks for the environment variable hint: I used `SSL_CERT_FILE` which should be simpler to use platform independently.

Happy easter, everyone!

m8mble:
I regret having mentioned platform independence ;). 

Unfortunately, I don't have a windows box at hand to reproduce. But I'll try my best to make things work using your CI...

m8mble:
Ok, at this point I could use some help. Whats so special about the VS12 pipeline, that makes it succeed while all others fail? Any ideas?

mpenick:
Thanks for test and trying to make it work on all platforms. I'll take a look.

mpenick:
I worked on this a bit yesterday on my local Windows machine. I thought I found the issue, and I'm able to get it working locally, but when I push it only one out of six CI builds works. I'll have to dig into this more.

mpenick:
Note: It works locally with many different version of VS.

tests/src/unit/tests/test_connection.cpp
```diff
@@ -35,14 +35,25 @@ using namespace datastax::internal::core;
...
 #endif
 }
 
+String current_dir()
+{
+  char buffer[256];
+#ifdef _WIN32
+  _getcwd(buffer, 256);
+#else
+  getcwd(buffer, 256);
```
bovine, line 52: Should ideally use sizeof(buffer) for both of these later references, to avoid having to replicate a constant.
```suggestion
  _getcwd(buffer, sizeof(buffer));
#else
  getcwd(buffer, sizeof(buffer));
```

tests/src/unit/tests/test_connection.cpp
```diff
@@ -190,6 +216,60 @@ TEST_F(ConnectionUnitTest, Ssl) {
...
+  const String cwd = current_dir();
+
+  // Generate certificate as file (which is used by our mock cluster) and import it
+  std::ofstream cert_buffer(cert_path.c_str());
+  cert_buffer << cert;
+  cert_buffer.close();
+  setenv("SSL_CERT_FILE", cert_path.c_str());
+  setenv("SSL_CERT_DIR", cwd.c_str());
+  std::cout << "Debug SslDefaultVerifyPaths: SSL_CERT_FILE " << cert_path << " " << cert << std::endl;
+  for (const auto var: {"SSL_CERT_FILE", "SSL_CERT_DIR"}) {
```
bovine, line 251: This newer style of "range-based for" doesn't seem to be supported on very old versions like MSVC10. On VC11 it seems to be failing on the inline definition of the array. Maybe try something like this?
```suggestion
  const char *sslvarnames[] = {"SSL_CERT_FILE", "SSL_CERT_DIR", nullptr};
  for (int i = 0; sslvarnames[i]; i++) {
    const char *var = sslvarnames[i];
```

PR #475, opened 2020-04-02
Reach full compatibility with OpenSSL 1.1

SeverinLeonhardt:
Despite what the documentation[1] says the code is not fully compatible
with OpenSSL 1.1 because it relies on compatibility wrappers behind
OPENSSL_API_COMPAT.

The OpenSSL documentation on library initialization[2] explains that
none of the startup fucntions are necessary starting with 1.1. The
changelog for "changes between 1.0.2h and 1.1.0 [25 Aug 2016]"[3]
mentions several removed shutdown functions. The removal of
`CRYPTO_set_locking_callback` is mentioned in a GitHub comment[4].

[1] https://docs.datastax.com/en/developer/cpp-driver/2.15/topics/building/
[2] https://wiki.openssl.org/index.php/Library_Initialization
[3] https://www.openssl.org/news/changelog.html
[4] https://github.com/openssl/openssl/issues/1260#issuecomment-228846202

### DIFF
--- a/src/ssl/ssl_openssl_impl.cpp
+++ b/src/ssl/ssl_openssl_impl.cpp
@@ -620,9 +620,11 @@ void free(void* ptr, const char* file, int line) { Memory::free(ptr); }
 void OpenSslContextFactory::internal_init() {
   CRYPTO_set_mem_functions(openssl::malloc, openssl::realloc, openssl::free);
 
+ #if OPENSSL_VERSION_NUMBER < 0x10100000L
   SSL_library_init();
   SSL_load_error_strings();
   OpenSSL_add_all_algorithms();
+  #endif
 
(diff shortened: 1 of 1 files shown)


src/ssl/ssl_openssl_impl.cpp
```diff
@@ -620,9 +620,11 @@ void free(void* ptr, const char* file, int line) { Memory::free(ptr); }
 void OpenSslContextFactory::internal_init() {
   CRYPTO_set_mem_functions(openssl::malloc, openssl::realloc, openssl::free);
 
+ #if OPENSSL_VERSION_NUMBER < 0x10100000L
```
mpenick, line 623: Maybe move these init. functions into the #ifdef block below (starting at line 629)?

mpenick: Maybe not. I don't think this will work for libressl.

mpenick:
Thanks for explanation and for fixing the issue.

PR #476, opened 2020-04-07
CPP-917 Fix infinite loop in token map calculation when using SimpleS…

mpenick:
…trategy


andrei-pavel:
I've experienced this bug, I would like this in the master branch if there's no hold up.

keef-cognitiv:
Even further, would be nice to have a release with this since 2.15.1 is entirely unusable for a SimpleStrategy cluster without it.

PR #482, opened 2020-08-06
Avoid potential linker error for http_message_needs_eof

debashish-ghosh:
Using other libraries having the same http_parser shows linker error.
> multiple definition of `http_message_needs_eof'


mpenick:
Thanks.

PR #484, opened 2020-09-25
Honor the value_length param

gwik:
Honor the value_length parameter when binding strings by
name in cass_statement_bind_string_by_name so that it can
work properly with non null-terminated strings.


gwik:
https://github.com/Metaswitch/cassandra-rs/issues/86

mpenick (approved):

jhgg:
Hi @mpenick  - would it be possible to release a point release for this fix soon? Thank you!

PR #490, opened 2021-01-18
Add tsan annotation for atomic_thread_fence

tavplubix:
Thread sanitizer does not support `atomic_thread_fence` with release/acquire order, so using of `atomic_thread_fence(MEMORY_ORDER_ACQUIRE)` in `RefCounted::dec_ref()` causes false positive reports like this: https://gist.github.com/alesapin/96b14caa274a291f12dcf84cddc8ff5c

Explicit annotation avoids such false positives.

### DIFF
--- a/src/atomic.hpp
+++ b/src/atomic.hpp
@@ -19,6 +19,24 @@
 
 #include "driver_config.hpp"
 
+#if !defined(THREAD_SANITIZER)
+#if defined(__has_feature)
+#if __has_feature(thread_sanitizer)
+#define THREAD_SANITIZER 1
+#endif
+#elif defined(__SANITIZE_THREAD__)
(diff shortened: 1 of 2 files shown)


tavplubix:
Btw, there is another data race which doesn't look false positive: https://gist.github.com/alesapin/bcb6ad9cd99e0c2e198ea4348db38145
I've fixed it with `std::shared_ptr` in our fork (https://github.com/ClickHouse-Extras/cpp-driver/commit/9cbc1a806df5d40fddbf84533b9873542c6513d8), but I'm not familiar with internals of the library, so there must be a better way to fix it.

mpenick:
Thanks. I'm running it through CI. Could you run a `make format` (that breaks our build)?

PR #492, opened 2021-02-09
Fix build with clang-{10,11}

vhavel:
Newer clang versions report several "Wimplicit-int-float-conversion"
warnings. In this commit, implicit int-to-float conversions are turned
into explicit static_casts in local code; violations in third-party
headers are suppressed by compiler pragma.


mpenick:
It looks like this breaks all our other gcc builds. :)

vhavel:
Sorry about that. I thought the second CI check simply isn't working as it runs on internal domain, now after a fix I see its actually green :)

I realized the previous approach is maybe to complicated given the driver is in maintenance mode, so I decided to just add this kind of warning to the list of disabled warnings for clang.

mpenick:
Thanks!

PR #493, opened 2021-03-04
Fixes for CPP-928, and trusted certs

maxdymond:
This PR consists of two commits:

---

[CPP-928](https://datastax-oss.atlassian.net/browse/CPP-928) Ensure server name information flows through from contact point configuration

Previously during Address name resolution, the server name information for a given Address
was lost. This fix ensures that the server name information flows through during the name
resolution process for a given Address.

---

Iterate over all certificates in a trusted cert BIO, not just the first

Previously the code which loaded a trusted certificate from file only
assumed that there was a single certificate in that file, meaning that 
using a certificate bundle for certificate verification would not work.

This fix allows the driver to read multiple trusted certificates out
of a BIO and provision them in the trusted certificate store.

---

Please let me know if you have any comments!

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -4602,6 +4602,12 @@ cass_ssl_add_trusted_cert_n(CassSsl* ssl,
  * common name or one of its subject alternative names. This implies the
  * certificate is also present. Hostname resolution must also be enabled.
  *
+ * Notes:
+ * - CASS_SSL_VERIFY_PEER_IDENTITY and CASS_SSL_VERIFY_PEER_IDENTITY_DNS are
+ *   mutually exclusive options.
+ * - The certificate Common Name is only checked against the IP address or
+ *   hostname if there are no Subject Alternative Names in the certificate.
+ *
  * <b>Default:</b> CASS_SSL_VERIFY_PEER_CERT
  *
  * @public @memberof CassSsl
--- a/src/ssl/ssl_openssl_impl.cpp
+++ b/src/ssl/ssl_openssl_impl.cpp
@@ -489,8 +489,8 @@ void OpenSslSession::verify() {
         return;
     }
   } else if (verify_flags_ &
-             CASS_SSL_VERIFY_PEER_IDENTITY_DNS) { // Match using hostnames (including wildcards)
-    switch (OpenSslVerifyIdentity::match_dns(peer_cert, hostname_)) {
+             CASS_SSL_VERIFY_PEER_IDENTITY_DNS) { // Match using the server name (including wildcards)
+    switch (OpenSslVerifyIdentity::match_dns(peer_cert, sni_server_name_)) {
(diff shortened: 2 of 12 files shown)


mpenick:
Hey, I think domain name verification is working as intended. The driver uses reverse DNS to lookup the host name for certificate validation (which is a side effect of Cassandra using IP addresses internally). We have some [SSL documentation](https://docs.datastax.com/en/developer/cpp-driver/2.15/topics/security/ssl/) that  make the use of those flags a bit clearer. I agree the header docs could do a better job of explain those.

> Cassandra uses IP addresses internally so those can be used directly for verification or a domain name can be used via reverse DNS (PTR record).

```
CassSsl* ssl = cass_ssl_new();

// CASS_SSL_VERIFY_PEER_IDENTITY_DNS (domain name)
cass_ssl_set_verify_flags(ssl, CASS_SSL_VERIFY_PEER_CERT | CASS_SSL_VERIFY_PEER_IDENTITY_DNS);

CassCluster* cluster = cass_cluster_new();

// Enable reverse DNS
cass_cluster_set_use_hostname_resolution(cluster, cass_true);

// ...
```

include/cassandra.h
```diff
@@ -4602,6 +4602,12 @@ cass_ssl_add_trusted_cert_n(CassSsl* ssl,
  * common name or one of its subject alternative names. This implies the
  * certificate is also present. Hostname resolution must also be enabled.
  *
+ * Notes:
+ * - CASS_SSL_VERIFY_PEER_IDENTITY and CASS_SSL_VERIFY_PEER_IDENTITY_DNS are
+ *   mutually exclusive options.
+ * - The certificate Common Name is only checked against the IP address or
```
mpenick, line 4608: > The certificate Common Name is only checked against the IP address or hostname if there are no Subject Alternative Names in the certificate.

Maybe this is a bug. If it's present in a SAN or the CN the it should probably be valid? I'd need to dig into that more.

src/ssl/ssl_openssl_impl.cpp
```diff
@@ -556,13 +540,30 @@ SslSession* OpenSslContext::create_session(const Address& address, const String&
 }
 
 CassError OpenSslContext::add_trusted_cert(const char* cert, size_t cert_length) {
-  X509* x509 = load_cert(cert, cert_length);
-  if (x509 == NULL) {
+  BIO* bio = BIO_new_mem_buf(const_cast<char*>(cert), cert_length);
```
mpenick, line 543: +1 on loading the whole cert chain. Awesome!

kw217:
@mpenick @maxdymond Happy New Year! I've pulled out the uncontroversial part of this PR (the change to read all certs) into a separate PR; @mpenick please can you re-review and merge? It's just a cherry-pick, no code changes since you reviewed above. Thanks.

PR #502, opened 2021-03-29
CPP-935: Fix latency-aware policy

mpenick:
Latency-aware policy update timer wasn't started on request processors.

### DIFF
--- a/src/latency_aware_policy.cpp
+++ b/src/latency_aware_policy.cpp
@@ -102,6 +102,8 @@ Host::Ptr LatencyAwarePolicy::LatencyAwareQueryPlan::compute_next() {
       return host;
     }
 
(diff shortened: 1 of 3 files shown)


mpenick:
Also, tested manually by running a query against a three node cluster then causing an artificial delay using traffic control:

```
tc qdisc add dev lo root handle 1: prio
tc qdisc add dev lo parent 1:3 handle 30: netem delay 1ms 
tc filter add dev lo protocol ip parent 1:0 prio 3 u32 \
   match ip dst 127.0.0.2 flowid 1:3
```

Watch logs for:

```
1617039588.003 [TRACE] (latency_aware_policy.cpp:105:virtual datastax::internal::core::Host::Ptr datastax::internal::core::LatencyAwarePolicy::LatencyAwareQueryPlan::compute_next()): Skipping 127.0.0.2 because latency is too high 4.312233
```

PR #503, opened 2021-03-30
Fix case sensitive identifiers, support quoted&unquoted user provided values

zakalibit:
Add support for quoted&unquoted user provided keyspaces values, the quoting is automatically handled in the driver.
This makes cpp driver interface usage cleaner, i.e. client code does not have to care about quoting case sensitive keyspace names.

Reason for this fix is, that our backend is passing unquoted keyspace values, as well we had a local fix in connector.cpp similar to what was in  pooled_connection.cpp before the CPP-747 fix, so this is a fix to handle both scenarios.

I have modified quote_id() to cater only for alphanumeric characters and underscores as per https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/valid_literal_r.html, i.e. there is no need to handle quotes in the middle of a string, which would be an invalid identifier. Have updated unit-tests as well.

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -178,7 +178,7 @@ void Connection::maybe_set_keyspace(ResponseMessage* response) {
   if (response->opcode() == CQL_OPCODE_RESULT) {
     ResultResponse* result = static_cast<ResultResponse*>(response->response_body().get());
     if (result->kind() == CASS_RESULT_KIND_SET_KEYSPACE) {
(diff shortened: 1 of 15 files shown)


zakalibit:
Can someone provide me with build failure reason, it compiles fine for me, tried all possible cmake configurations.

zakalibit:
Ahh I see some integration tests failed, will look in to that

zakalibit:
I have come up with a smaller change and created a new pull request.

PR #504, opened 2021-04-01
Handle unquoted case sensitive keyspaces

zakalibit:
Add support for quoted&unquoted user provided keyspaces values, the quoting is automatically handled in the driver. This makes cpp driver interface usage cleaner, i.e. client code does not have to care about quoting case sensitive keyspace names.

Reason for this fix is, that our backend is passing unquoted keyspace values, as well we had a local fix in connector.cpp similar to what was in pooled_connection.cpp before the CPP-747 fix, so this is a fix to handle both scenarios.

I have modified quote_id() to cater only for alphanumeric characters and underscores as per https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/valid_literal_r.html, i.e. there is no need to handle quotes in the middle of a string, which would be an invalid identifier. Have updated unit-tests as well.

This makes **cass_session_connect_keyspace**() accept quoted and unquoted keyspace names.

I have looked as well at:
**cass_data_type_set_keyspace**
**cass_batch_set_keyspace**
**cass_statement_set_keyspace**

but it does not seem that they require any changes

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -101,7 +101,7 @@ String implode(const Vector<String>& vec, const char delimiter /* = ' ' */) {
   return str;
 }
 
-bool not_isspace(int c) { return !::isspace(c); }
+static bool not_isspace(int c) { return !::isspace(c); }
 
 String& trim(String& str) {
   // Trim front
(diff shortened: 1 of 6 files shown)


zakalibit:
@mpenick not sure why jenkins is failing on this one, could you give me a hint

mpenick:
`std::find_if_not()` is C++11. It fails on centos 6's default gcc.
```
[2021-04-07T06:19:26.534Z] /home/jenkins/workspace/drivers_cpp_oss_PR-504/src/utils.cpp:120: error: ‘find_if_not’ is not a member of ‘std’
```

zakalibit:
> `std::find_if_not()` is C++11. It fails on centos 6's default gcc.
> 
> ```
> [2021-04-07T06:19:26.534Z] /home/jenkins/workspace/drivers_cpp_oss_PR-504/src/utils.cpp:120: error: ‘find_if_not’ is not a member of ‘std’
> ```

I was wandering if you really need to support anything lower than c++11? could use move semantics for more perf gain

andrei-pavel:
CentOS 6 is EOL since November 2020.

mpenick:
> I was wandering if you really need to support anything lower than c++11? could use move semantics for more perf gain

zakalibit:
Yes, I'll fix this one anyways, it is a simple fix

zakalibit:
done

zakalibit:
> 
> If it's not too much effort do you want to get this PR in first supporting pre-C++11?
> 
done

src/utils.cpp
```diff
@@ -112,39 +112,28 @@ String& trim(String& str) {
 }
 
 static bool is_lowercase(const String& str) {
-  if (str.empty()) return true;
-
-  char c = str[0];
-  if (!(c >= 'a' && c <= 'z')) return false;
+  return str.end() == std::find_if(str.begin(), str.end(), ::isupper);
```
mpenick, line 115: This changes the semantics. `is_lowercase()` is a misnomer here it's more like "is this a valid lowercase CQL identifier?" that doesn't need to be quoted.  `::isupper` would let invalid chars through e.g. `-`, `.`. Also, `_` can't be the first char.

zakalibit: Well, arguably this function does not have to worry about if it is a valid cql identifier, cassandra will do that validation, i.e. all we do here is check if any upper case chars, then will add quotes.

And if it does check for the above, what it suppose to do about it?

Does the driver validate the whole CQL statement? probably not.

zakalibit: It can be reverted I think, but I do not see it making any real difference, apart from being overcomplicated with no reason.

zakalibit: @mpenick any thoughts, do you want this reverted?

PR #505, opened 2021-04-02
Performance improvement in Decoder::decode_value

zakalibit:
Please consider this optimisation, it gives about 39% improvement in **decode_value**, this results in 2-3% improvement in our production backend, which is considerable taking in account the scale of our deployments.

**decode_value** 7.55%, on the right side **Value::operator=** & **Value::~Value**
Screenshot 2021-04-02 at 12 35 39

**decode_value** 4.67%
Screenshot 2021-04-02 at 12 49 19

### DIFF
--- a/src/user_type_field_iterator.cpp
+++ b/src/user_type_field_iterator.cpp
@@ -25,5 +25,6 @@ bool UserTypeFieldIterator::next() {
     return false;
   }
   current_ = next_++;
-  return decoder_.decode_value(current_->type, value_);
+  value_ = decoder_.decode_value(current_->type);
+  return !!value_.data_type();
 }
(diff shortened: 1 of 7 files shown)


mpenick:
Nice improvement! Thanks. Could you run `make format` so this will make it through CI?

zakalibit:
> Nice improvement! Thanks. Could you run `make format` so this will make it through CI?

done

src/user_type_field_iterator.cpp
```diff
@@ -25,5 +25,6 @@ bool UserTypeFieldIterator::next() {
     return false;
   }
   current_ = next_++;
-  return decoder_.decode_value(current_->type, value_);
+  value_ = decoder_.decode_value(current_->type);
+  return !!value_.data_type();
```
mpenick, line 29: What do you think about using the following for these checks?

```cpp
return value.is_valid();
```

where `is_valid()` is defined as:
```cpp
bool Value::is_valid() {
   return value_.data_type();
}
```

zakalibit: looks good, it is more intuitive, will do the change

zakalibit: done

PR #508, opened 2021-04-27
Performance improvement reduce Value DataType refcounting

zakalibit:
Value object DataType refcounting it is quite a perf hog, this optimisation reduces decode_value() by ~72%, it saves ~20-24% cpu in our backend.

In the shown code path is ~82% saving

decode_value() - 12.8%
Screenshot 2021-04-27 at 14 51 43

decode_value() - 2.29%
Screenshot 2021-04-27 at 14 34 47

Had to change CollectionIterator::decode_value() as it was creating a temporary, DataType::ConstPtr creates a copy, it seems a strange and dangerous behaviour.

I understand that the solution is not pretty, but being unable to use at least c++11, it is the best I could come up with.

### DIFF
--- a/src/collection_iterator.cpp
+++ b/src/collection_iterator.cpp
@@ -29,7 +29,7 @@ bool CollectionIterator::next() {
 bool CollectionIterator::decode_value() {
   if (collection_->value_type() == CASS_VALUE_TYPE_MAP) {
     const DataType::ConstPtr& data_type =
-        ((index_ & 1) == 0) ? collection_->primary_data_type() : collection_->secondary_data_type();
+        (index_ % 2 == 0) ? collection_->primary_data_type() : collection_->secondary_data_type();
     value_ = decoder_.decode_value(data_type);
   } else {
     value_ = decoder_.decode_value(collection_->primary_data_type());
--- a/src/collection_iterator.cpp
+++ b/src/collection_iterator.cpp
@@ -27,15 +27,14 @@ bool CollectionIterator::next() {
 }
 
 bool CollectionIterator::decode_value() {
-  DataType::ConstPtr data_type;
   if (collection_->value_type() == CASS_VALUE_TYPE_MAP) {
-    data_type =
-        (index_ % 2 == 0) ? collection_->primary_data_type() : collection_->secondary_data_type();
+    const DataType::ConstPtr& data_type =
+        ((index_ & 1) == 0) ? collection_->primary_data_type() : collection_->secondary_data_type();
+    value_ = decoder_.decode_value(data_type);
   } else {
-    data_type = collection_->primary_data_type();
+    value_ = decoder_.decode_value(collection_->primary_data_type());
   }
 
-  value_ = decoder_.decode_value(data_type);
   return value_.is_valid();
 }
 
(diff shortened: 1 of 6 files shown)


accelerated:
Although I won't deny that *not* incrementing/decrementing atomic refcounts will improve your performance, you are trading safety for performance. Typically it's considered BAD practice to pass smart pointers by reference, unless you absolutely, positively know that refcount cannot go to zero while you're storing it or using it. I'll give you an example: You take a reference to a shared pointer, the CPU does a context switch on you, and the new execution thread (whatever that happens to be), executes some code which decrements your shared pointer. Assume it's now 0 and the shared resource has been freed. Another context switch and you're back to your original thread where you have this reference on the stack. You now access this resource and kaboom. Debugging this crash will also not be easy...
I'm not saying this can happen in this particular example, but it illustrates why it's a dangerous practice. If you need to improve performance in the hot path where this shared pointer may be passed around lots of times like in a tight loop or something, then perhaps it's better to get it by value somewhere outside the loop or hot path and then pass it by reference in the call chain following that. That way you reduce your atomic operations while guaranteeing resource lifetime.

zakalibit (commented):
> Although I won't deny that _not_ incrementing/decrementing atomic refcounts will improve your performance, you are trading safety for performance. Typically it's considered BAD practice to pass smart pointers by reference, unless you absolutely, positively know that refcount cannot go to zero while you're storing it or using it. I'll give you an example: You take a reference to a shared pointer, the CPU does a context switch on you, and the new execution thread (whatever that happens to be), executes some code which decrements your shared pointer. Assume it's now 0 and the shared resource has been freed. Another context switch and you're back to your original thread where you have this reference on the stack. You now access this resource and kaboom. Debugging this crash will also not be easy...
> I'm not saying this can happen in this particular example, but it illustrates why it's a dangerous practice. If you need to improve performance in the hot path where this shared pointer may be passed around lots of times like in a tight loop or something, then perhaps it's better to get it by value somewhere outside the loop or hot path and then pass it by reference in the call chain following that. That way you reduce your atomic operations while guaranteeing resource lifetime.

I do not  disagree with above, but that is general theory, however smart pointers for DataType beyond  result object returned by the future, is excessive, all DataTypes objects live in a metadata class member that holds them as smart pointers, then I hardly see how anything that is done with that result object can and should happen on different threads, even if it does, all iterators, rows, values have DataType objects lifetime warrantied by the result object lifetime, so all that refcounting in this context is a waste of cpu.

accelerated:
Perhaps it's best to use unique pointers rather than shared pointers, if the whole use-case is to ensure auto-deletion of the resource?

zakalibit:
> Perhaps it's best to use unique pointers rather than shared pointers, if the whole use-case is to ensure auto-deletion of the resource?

First is good to understand the data flow, then decide if unique, or shred pointers are required at all.
One and most common flow from what I understand, is:
- network thread creates the response object and populates metadata
- then it hand's over response object to the caller, through a callback, or future object
- the caller is consuming the data, then releases the response object and the rest of the data owned by it


So it seems that at any given point in time there is a single owner/consumer for the response object and metadata, so this could be implemented without any smart pointers just store DataType objects on the response object by value, maybe that was expensive before, but should not be the case with move semantics since c++11

PR #510, opened 2021-05-11
Row values decoding performance optimisation

zakalibit:
This is a "better" row values decoding optimisation compared to the first attempt https://github.com/datastax/cpp-driver/pull/508

It skips creating **Value** objects and populating the row vector on every next row decode, instead it simply updates **Value's** **Decoder** instance and **Value** internal state.
So there is no refcounting, no vector cleanup and memory allocations.

**cass::ResultIterator::next()** - 13.28%
Screenshot 2021-05-11 at 11 11 15

**cass::ResultIterator::next()** - 4.45%
Screenshot 2021-05-11 at 11 08 55

### DIFF
--- a/src/value.cpp
+++ b/src/value.cpp
@@ -214,6 +214,12 @@ bool Value::update(const Decoder& decoder) {
   if (!decoder_.is_null()) {
     if (data_type_->is_collection()) {
       return decoder_.decode_int32(count_);
+    } else if (data_type->is_tuple()) {
+      const CompositeType& composite_type = static_cast<const CompositeType&>(*data_type);
+      count_ = composite_type.types().size();
+    } else if (data_type->is_user_type()) {
+      const UserType& user_type = static_cast<const UserType&>(*data_type);
+      count_ = user_type.fields().size();
     }
   } else {
     count_ = 0;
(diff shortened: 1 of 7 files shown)


zakalibit:
@mpenick could you please let me know if this PR has a chance to be merged.

src/value.cpp
```diff
@@ -209,6 +209,19 @@ Value::Value(const DataType::ConstPtr& data_type, Decoder decoder)
   }
 }
 
+bool Value::update(const Decoder& decoder) {
+  decoder_ = decoder;
+  if (!decoder_.is_null()) {
+    if (data_type_->is_collection()) {
```
mpenick, line 215: This also needs to handle tuple and UDT cases.

zakalibit: Ahh indeed as it resets the count on a null value

zakalibit: Fixed, I think there is an intermittent unittest failure

zakalibit: @mpenick could you please review the fix

mpenick: I'm reviewing and running through a couple integration tests.

PR #511, opened 2021-06-30
CPP-944: Fix test builds on aarch64

wash-amzn:
The signed-ness of 'char' varies by platform. Fix tests that are constructing binary input data as 'char' arrays with negative literals by switching them over to 'signed char' arrays.

https://datastax-oss.atlassian.net/browse/CPP-944

PR #512, opened 2021-06-30
CPP-943: Fix intermittent failures of SchemaAgreementUnitTest.Simple

wash-amzn:
The test stops cluster nodes and then performs queries. It is possible for that query to start being executed before the session detects the connection going down, resulting in test failures.

https://datastax-oss.atlassian.net/browse/CPP-943


mpenick:
This is also a code format issue. Please run `make format`.

wash-amzn:
As with my other PR (https://github.com/datastax/cpp-driver/pull/511), the formatter wanted to change a file I didn't modify, so however you wish to proceed with that other one would apply here as well.

PR #513, opened 2021-08-10
Fix SNI events

mpenick:
When using SNI routing the driver is not mapping event address to the representation used by the driver (an `Address` with an IP and SNI name). It needs to be mapped using the `rpc_address` from the system tables.

### DIFF
--- a/src/control_connection.cpp
+++ b/src/control_connection.cpp
@@ -307,7 +307,7 @@ static NopControlConnectionListener nop_listener__;
 ControlConnectionSettings::ControlConnectionSettings()
     : use_schema(CASS_DEFAULT_USE_SCHEMA)
     , use_token_aware_routing(CASS_DEFAULT_USE_TOKEN_AWARE_ROUTING)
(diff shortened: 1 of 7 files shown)


src/control_connection.cpp
```diff
@@ -395,12 +395,10 @@ void ControlConnection::handle_refresh_node(RefreshNodeCallback* callback) {
   const Row* row = NULL;
   ResultIterator rows(callback->result().get());
 
-  while (rows.next() && !found_host) {
+  while (!found_host && rows.next()) {
     row = rows.row();
     if (callback->is_all_peers) {
```
zakalibit, line 400: would `callback->is_all_peers` not be a constant for the duration fo the while loop? Maybe we could have something like
`
bool found_host = callback->is_all_peers && callback->result().row_count() > 0;
while (!found_host && rows.next()) {
  row = rows.row();
  found_host = settings_.address_factory->is_peer(row, connection_->host(), callback->address);
}
`

mpenick: Nice suggestion.

PR #515, opened 2021-10-06
Log instead of assert in order to avoid core dumps on shutdown

absurdfarce:
Customer complained that core dumps from an unclean event loop were interfering with their regular evaluation of core dumps.  The code in question occurs at event loop shutdown so the extra handles shouldn't be much of a concern.

### DIFF
--- a/src/event_loop.cpp
+++ b/src/event_loop.cpp
@@ -61,7 +61,8 @@ EventLoop::~EventLoop() {
       uv_run(loop(), UV_RUN_DEFAULT);
       rc = uv_loop_close(loop());
       if (rc != 0) {
-        LOG_INFO("Event loop still has pending handles");
+        uv_print_all_handles(loop(), stderr);
+        LOG_ERROR("Event loop still has pending handles");
       }
     }
   }


src/event_loop.cpp
```diff
@@ -61,8 +61,7 @@ EventLoop::~EventLoop() {
       uv_run(loop(), UV_RUN_DEFAULT);
       rc = uv_loop_close(loop());
       if (rc != 0) {
-        uv_print_all_handles(loop(), stderr);
-        assert(false && "Event loop still has pending handles");
+        LOG_INFO("Event loop still has pending handles");
```
absurdfarce, line 64: This is the brain-dead simple version of the fix, but unfortunately we lose out on the uv_print_all_handles output here.  We _could_ work around this fmemopen + passing the resulting stream to uv_print_all_handles + making that the log message, but that requires a good idea of the size of the message to be generated.  And that isn't easy to obtain; best I've been able to come up with is uv_walk to count handles * an estimated per-handle message size + some slop... and that's not great.

I'm happy to pursue this if we believe the list of handles is useful in a log message but it seemed best to start with the super-simple version in case this was adequate.

zakalibit: I think we can try a cleaner event loop shutdown, something along these lines
`
rc = uv_loop_close(loop());
if (rc == UV_EBUSY) {
    uv_stop(loop());
    uv_walk(loop(), [](uv_handle_t* handle, void* arg) {
			             if (uv_is_closing(handle) == 0) {
				         uv_close(handle, [](uv_handle_t* h){});
				    }}, nullptr);
    uv_run(loop(), UV_RUN_DEFAULT);
}
`
or that can be incapsulated in a while loop, maybe with retry count so it does not spin endlessly 

`
while(UV_EBUSY == uv_loop_close(loop())) {
    ....
}
`

absurdfarce:
File https://datastax-oss.atlassian.net/browse/CPP-946 in order to provide some additional context

src/event_loop.cpp
zakalibit: Forgot that c++11 can not be used yet, or it can?
Otherwise 2 more static methods on `EventLoop::` is needed
```
void EventLoop::on_uv_close(uv_handle_t *handle) {
  return;
}
void EventLoop::on_uv_close_walk(uv_handle_t *handle, void *arg) {
  uv_close(handle, on_uv_close);
}
```
so then
```
rc = uv_loop_close(loop());

if (rc == UV_EBUSY) {
    uv_stop(loop());
    uv_walk(loop(), on_uv_close_walk, nullptr);
    uv_run(loop(), UV_RUN_NOWAIT);

    rc = uv_loop_close(loop());
    if (rc == UV_EBUSY) {
       LOG_INFO("Event loop still has pending handles");
       uv_print_all_handles(loop(), stderr);
    }
}
```
maybe some extra logging, or some debug logging for troubleshooting, so if there is still issues in this area, extra information can be provided by the users simply by enabling debug

src/event_loop.cpp
```diff
@@ -62,7 +62,7 @@ EventLoop::~EventLoop() {
       rc = uv_loop_close(loop());
       if (rc != 0) {
         uv_print_all_handles(loop(), stderr);
-        assert(false && "Event loop still has pending handles");
+        LOG_ERROR("Event loop still has pending handles");
```
absurdfarce, line 65: This winds up printing "Event loop still has pending handles" in a (potentially) different location than the dump of handles to stderr.  Should we change the verbiage here to something like "Event loop still has pending handles; see status of event loop handles on stderr"?

src/event_loop.cpp
absurdfarce: For now I'd like to keep this focused on the removal of the core dump process as that has an immediate impact on a customer problem.  @zakalibit I've filed https://datastax-oss.atlassian.net/browse/CPP-947 to consider stronger event loop shutdown logic (with a pointer back to this PR).  Please feel free to add whatever additional context you'd like on that ticket!

absurdfarce:
Closing as no longer relevant given https://github.com/datastax/cpp-driver/pull/517

PR #518, opened 2021-10-21
Replace deprecated function for OpenSSL >= 3.0

pjgeorg:
ERR_get_error_line_data is deprecated in OpenSSL >= 3.0.
It can be replaced by the newer function ERR_get_error_all added in OpenSSL 3.0.


remicollet:
Please merge, Fedora 36, RHEL 9 Beta and CentOS 9 Stream now use OpenSSL 3.0

PR #522, opened 2022-01-10
Iterate over all certificates in a trusted cert BIO, not just the first

kw217:
Previously the code which loaded a trusted certificate from file only assumed that there was a single certificate in that file, meaning that using a certificate bundle for certificate verification would not work.

This fix allows the driver to read multiple trusted certificates out of a BIO and provision them in the trusted certificate store.

_This code was already reviewed in #493 but that PR contained other fixes as well and has not yet been merged. Instead, this PR pulls that out separately. It is also now based on master (2.16.2)._

### DIFF
--- a/src/ssl/ssl_openssl_impl.cpp
+++ b/src/ssl/ssl_openssl_impl.cpp
@@ -240,22 +240,6 @@ end:
   return ret;
 }
 
-static X509* load_cert(const char* cert, size_t cert_size) {
-  BIO* bio = BIO_new_mem_buf(const_cast<char*>(cert), cert_size);
-  if (bio == NULL) {
-    return NULL;
-  }
-
-  X509* x509 = PEM_read_bio_X509(bio, NULL, pem_password_callback, NULL);
-  if (x509 == NULL) {
-    ssl_log_errors("Unable to load certificate");
-  }
-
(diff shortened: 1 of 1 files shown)


kw217:
Looks like it's passing appveyor now - apologies, my first push had an issue that was caught by the UTs (turns out you have to call `ERR_get_error()` even if you don't use it, otherwise the next SSL function fails mysteriously).

This should now be ready to review.

absurdfarce:
@kw217 Thanks for the contribution!  And my sincere apologies for the lengthy delay in getting to this review.

I very much like the idea of pulling these changes out of the previous PR and managing them here.  They're very self-contained (which this PR makes clear) and the implementation here is clean so I'm happy to merge this as it stands.

Regarding the test failures: the Jenkins build points to an internal DataStax Jenkins server which isn't available publicly.  The good news is that the failures noted on that Jenkins run are due to a very minor formatting issue that only seems to occur on one of our test platforms.  I'll include a fix for that issue in a short follow-up commit.

With that in mind, I believe we have what we need to merge this now.  Thanks again for the PR!

absurdfarce:
For sake of completeness the follow-up PR is [here](https://github.com/datastax/cpp-driver/pull/526).

PR #523, opened 2022-01-10
Deprecate `cass_cluster_set_use_hostname_resolution()`

mpenick:
Add documentation that MITM is possible when using
`CASS_SSL_VERIFY_PEER_IDENTITY_DNS` with
`cass_cluster_set_use_hostname_resolution()`

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -2565,8 +2565,8 @@ cass_cluster_set_use_schema(CassCluster* cluster,
 /**
  * Enable/Disable retrieving hostnames for IP addresses using reverse IP lookup.
  *
- * This is useful for authentication (Kerberos) or encryption (SSL) services
- * that require a valid hostname for verification.
+ * @deprecated Do not use. Using reverse DNS lookup to verify the certificate
+ * does not protect against man-in-the-middle attacks.
  *
  * <b>Default:</b> cass_false (disabled).
  *
@@ -2578,9 +2578,9 @@ cass_cluster_set_use_schema(CassCluster* cluster,
  *
  * @see cass_cluster_set_resolve_timeout()
  */
-CASS_EXPORT CassError
+CASS_EXPORT CASS_DEPRECATED(CassError
 cass_cluster_set_use_hostname_resolution(CassCluster* cluster,
-                                         cass_bool_t enabled);
+                                         cass_bool_t enabled));
 
 /**
  * Enable/Disable the randomization of the contact points list.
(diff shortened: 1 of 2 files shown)


include/cassandra.h
```diff
@@ -2565,8 +2565,8 @@ cass_cluster_set_use_schema(CassCluster* cluster,
 /**
  * Enable/Disable retrieving hostnames for IP addresses using reverse IP lookup.
  *
- * This is useful for authentication (Kerberos) or encryption (SSL) services
- * that require a valid hostname for verification.
+ * @deprecated Do not use. Using reverse DNS lookup to verify the certificate
+ * does not protect against man-in-the-middle attacks.
```
absurdfarce, line 2569: Perhaps include mention of CPP-942 inline in comments for additional context?

Although it doesn't look like there's much reference to CPP tickets in comments elsewhere in the source so maybe not.

include/cassandra.h
```diff
@@ -4598,9 +4598,9 @@ cass_ssl_add_trusted_cert_n(CassSsl* ssl,
  * CASS_SSL_VERIFY_PEER_IDENTITY - IP address matches the certificate's
  * common name or one of its subject alternative names. This implies the
  * certificate is also present.
- * CASS_SSL_VERIFY_PEER_IDENTITY_DNS - Hostname matches the certificate's
- * common name or one of its subject alternative names. This implies the
- * certificate is also present. Hostname resolution must also be enabled.
+ * CASS_SSL_VERIFY_PEER_IDENTITY_DNS -  Do not use. The requires the use of
+ * reverse DNS lookup which in not sufficient to protect against
```
absurdfarce, line 4602: s/in/is/

topics/security/ssl/README.md
```diff
@@ -172,24 +172,28 @@ If a unique certificate has been generated for each Cassandra node with the IP a
...
 
-// Or use: CASS_SSL_VERIFY_PEER_IDENTITY_DNS (domain name)
-cass_ssl_set_verify_flags(ssl, CASS_SSL_VERIFY_PEER_CERT | CASS_SSL_VERIFY_PEER_IDENTITY_DNS);
 ```
 
-If using a domain name to verify the peer's identity then hostname resolution
-(reverse DNS) needs to be enabled:
-
-**NOTE:** This is also disabled by default.
+**Important:** This section use to suggest using reverse DNS lookup as a way to validate the peer's certificate i.e. using `CASS_SSL_VERIFY_PEER_IDENTITY_DNS` with `cass_cluster_set_use_hostname_resolution(cluster, cass_true)`.  This is susceptible to man-in-the-middle (MITM) attacks and is no longer recommended.
```
absurdfarce, line 180: s/use/used/

absurdfarce (approved):
A couple minor typo nits but otherwise this looks pretty good

absurdfarce:
Something else just occurred to me a little bit ago: we probably still need a changelog mention for this even though it isn't actually changing (behaviour in) code, right?

absurdfarce:
Merging this in for now, will address the changelog update separately once we get closer to a release

PR #524, opened 2022-03-07
CPP-953 Naive attempt to avoid pure virtual function calls at cluster shutdown

absurdfarce:
Attempt to avoid entangling PrepareHostHandler and Cluster by avoiding executing host up checks when closing the handler

### DIFF
--- a/src/prepare_host_handler.cpp
+++ b/src/prepare_host_handler.cpp
@@ -67,8 +67,6 @@ void PrepareHostHandler::prepare(uv_loop_t* loop, const ConnectionSettings& sett
 }
 
 void PrepareHostHandler::on_close(Connection* connection) {
-  callback_(this);
-
   dec_ref(); // The event loop is done with this handler
 }
 


src/prepare_host_handler.cpp
```diff
@@ -67,8 +67,6 @@ void PrepareHostHandler::prepare(uv_loop_t* loop, const ConnectionSettings& sett
 }
 
 void PrepareHostHandler::on_close(Connection* connection) {
-  callback_(this);
-
```
absurdfarce, line 71: @mpenick What do you think of this idea?  The callback here is Cluster::on_prepare_host_up and it doesn't look like there's any benefit to calling it when we're shutting down the cluster anyway.  And if we don't call it we should avoid tangling up Cluster and PrepareHostHandler method calls... which should remove the possibility of a pure virtual function call.

Does this make sense or am I missing something?

tests/src/integration/tests/test_prepare_on.cpp
```diff
@@ -369,7 +369,7 @@ CASSANDRA_INTEGRATION_TEST_F(PrepareOnUpAndAddTests, NotPreparedOnUpWhenDisabled
 
   // Wait for the node to become available and verify no statements have been
   // prepared
-  wait_for_node_on_session(session, 1);
+  wait_for_node_on_session(session_for_node(1), 1);
```
absurdfarce, line 372: Minor test change.  wait_for_node_on_session() is querying system.local so to get a robust comparison we should always use a clean session, especially after bootstrapping a new node.  I was seeing these tests fail due to a belief that the node was unavailable; after digging in it looked as if the session was simply querying the wrong node.

src/prepare_host_handler.cpp
mpenick: I think this is going after the symptom instead of fixing the cause. My bet is the listener which is `SessionBase` is being freed before the the Cluster object is finished with it.

absurdfarce:
Replaced by #578

PR #527, opened 2022-03-28
CPP-956 Fix gcc compiler warnings for integration tests

absurdfarce:


tests/src/integration/objects/cluster.hpp
```diff
@@ -216,11 +216,18 @@ class Cluster : public Object<CassCluster, cass_cluster_free> {
...
+  #  pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+  #endif
   Cluster& with_hostname_resolution(bool enable = true) {
     EXPECT_EQ(CASS_OK, cass_cluster_set_use_hostname_resolution(
                            get(), (enable == true ? cass_true : cass_false)));
     return *this;
   }
+  #if defined(__GNUC__) || defined(__INTEL_COMPILER)
+  #  pragma GCC diagnostic pop
+  #endif
```
absurdfarce, line 230: We _might_ be able to do away with this function all together.  It's only used in some of the DSE tests and those tests appear to pass without them, at least against DSE 6.7.7.  The code above seems less intrusive than potentially changing how a test behaves so I decided to start with this... but I could be convinced otherwise.

absurdfarce:
"make" after "cmake -DCASS_BUILD_INTEGRATION_TESTS=On .." completes with no warnings after this change

tests/src/integration/integration.cpp
```diff
@@ -148,6 +148,10 @@ void Integration::SetUp() {
   data_center_nodes.push_back(number_dc2_nodes_);
 
   if (is_ccm_requested_) {
+#if defined(__GNUC__) || defined(__INTEL_COMPILER)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcatch-value"
```
mpenick, line 153: Can the code just be fixed to resolve the warning?

tests/src/integration/objects/cluster.hpp
```diff
@@ -206,21 +206,28 @@ class Cluster : public Object<CassCluster, cass_cluster_free> {
...
+ * This is useful for authentication (Kerberos) or encryption (SSL)
+ * services that require a valid hostname for verification.
+ *
+ * @param enable True if hostname resolution should be enabled; false
+ *               otherwise (default: true)
+ * @return Cluster object
+ */
+#if defined(__GNUC__) || defined(__INTEL_COMPILER)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
mpenick, line 221: I think it's fine to have deprecation warnings.

mpenick (changes requested):
I don't think we should be using pragma's for this...It's kind of dangerous.
* The first warning is valid and should be fixed.
* The other two are for our deprecated functions, the fix is to eventually remove those functions and we should see those warnings until they're removed.

absurdfarce:
Replaced by #577

PR #528, opened 2022-03-29
Fix with newer LibreSSL

gahr:
I'm not totally positive I picked the right version number.

### DIFF
--- a/src/ssl/ssl_openssl_impl.cpp
+++ b/src/ssl/ssl_openssl_impl.cpp
@@ -200,7 +200,8 @@ static int SSL_CTX_use_certificate_chain_bio(SSL_CTX* ctx, BIO* in) {
     int r;
     unsigned long err;
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2060000fL)
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || \
+    (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2090100fL)
     if (ctx->extra_certs != NULL) {
       sk_X509_pop_free(ctx->extra_certs, X509_free);
       ctx->extra_certs = NULL;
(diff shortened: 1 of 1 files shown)


absurdfarce:
Thanks for the contribution @gahr!

After a bit of code spelunking it looks like SSL_CTX_clear_chain_certs was introduced in 2.9.1 by way of [this commit](https://github.com/libressl-portable/openbsd/commit/07062ed31708ec955c12de39be204aa7169604b3).  If that's the case I believe the version number we want in our check would be "0x2090100fL" (based on the [opensslv.h in 2.9.1](https://github.com/libressl-portable/openbsd/blob/libressl-v2.9.1/src/lib/libcrypto/opensslv.h#L6)).  Are you able to verify that your application works correctly with that code in place?

gahr:
Hi @absurdfarce, yup that looks correct, thanks for double checking. What I wasn't able to properly figure out is when the struct became opaque. We might have an interval between when that happened and when the new API became available. Hopefully nobody is using a libressl version older then 2.9.1, which came out 3 years ago :)

So, yes, a check on 2.9.1 works fine to me.

src/ssl/ssl_openssl_impl.cpp
```diff
@@ -200,7 +200,7 @@ static int SSL_CTX_use_certificate_chain_bio(SSL_CTX* ctx, BIO* in) {
     int r;
     unsigned long err;
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2060000fL)
```
absurdfarce, line 203:
```suggestion
#if OPENSSL_VERSION_NUMBER < 0x10100000L || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2090100fL)
```

absurdfarce:
Makes sense @gahr.  I'm okay with the change since it certainly makes our LibreSSL support more correct than it was.  We can always refine this check if we see issues with LibreSSL versions in the interval you describe.

If you can apply the change I suggested I think we're almost there.  I'm looking into the issues around the build failure on Jenkins; I'll see if there's an easy fix there as well.

absurdfarce:
@gahr Jenkins build failures were just a formatting issue.  I've updated my suggested change to include a formatting fix which should address that issue as well.  If you're good with the changes there just accept them and we'll see if the resulting build is green again. :)

gahr:
Thanks!

absurdfarce:
Thank you @gahr!  Looks like we're all set here!

PR #530, opened 2022-05-06
CPP-959 Update README to include directions for finding post-2.16.0 releases

absurdfarce:


README.md
```diff
@@ -11,8 +11,11 @@ products:
 ## Getting the Driver
 
 Binary versions of the driver, available for multiple operating systems and
-multiple architectures, can be obtained from our [download server]. The
-source code is made available via [GitHub].
+multiple architectures, can be obtained from the appropriate download server.
+Releases up to and including 2.16.0 can be found on our
+[download server]. Subsequent releases can be found on our Artifactory
+instance; consult the release announcement on our [mailing list] for
+individual URLs. The source code is made available via [GitHub].
```
absurdfarce, line 18: I thought about basically copying the links from the release announcement(s) here and referencing them directly but that doesn't seem to scale at all.  We _could_ move that out of the way of the README and into some kind of RELEASES.md file which would contain that info... thoughts?

absurdfarce:
This is being addressed by similar (but slightly different) means in PR #552

PR #531, opened 2022-05-26
[datatype] fix equals method for udt datatypes across schema upgrades

jhgg:
We encountered a bug with this driver, where an altered UDT would cause `cass_statement_bind_user_type_by_name_n` to fail on a prepared statement that was prepared prior to the db altered.

I think that this should be a sensible fix, it for sure fixes our issue... however, I have no idea if this is actually sound!

### DIFF
--- a/src/data_type.hpp
+++ b/src/data_type.hpp
@@ -28,6 +28,8 @@
 #include "types.hpp"
 #include "vector.hpp"
 
+#include <algorithm>
+
 namespace datastax { namespace internal { namespace core {
 
 class Collection;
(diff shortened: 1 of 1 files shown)


jhgg:
Thinking about this more, I think that although this is semantically correct for `IsValidDataType`, it may not be right for `equals`  - so maybe we might introduce a separate method that just does the check of mutual fields and have is valid data type invoke that instead.

src/data_type.hpp
```diff
@@ -365,11 +367,15 @@ class UserType : public DataType {
...
-      return false;
-    }
 
-    for (size_t i = 0; i < fields_.size(); ++i) {
+    // UDT's can be considered equal as long as the mutual first fields shared
+    // between them are equal. UDT's are append only as far as fields go, so a
+    // newer 'version' of the UDT data type after a schema change event should be
+    // treated as equivalent in this scenario, by simply looking at the first N
+    // mutual fields they should share.
+    size_t min_fields_size = std::min(fields_.size(), user_type->fields_.size());
```
mpenick, line 376: This seems reasonable to me. It does seem that after creating a type that a type can only be altered in the following ways:

* Rename field: `ALTER TYPE address RENAME zip TO zipcode AND street_name TO street` 
* Add a field: `ALTER TYPE address ADD country text`

It'd be nice to have some tests that verify this is a correct assumption.

mpenick (commented):
Thanks for the PR.

jhgg:
> Thanks for the PR.

What do you think about my second comment? Should I instead refactor this into a "is valid data type for" method, that way equals still remains an equality check? 

Additionally, if that's the case, we can drop the *name* check on the field, this will allow us to handle renaming of a field, and still considering them equal.

src/data_type.hpp
jhgg: > It'd be nice to have some tests that verify this is a correct assumption.

I'm not entirely sure how we would test that? Are you suggesting we try to test that the database rejects invalid DDL? 

There is a third way, you can change the type of a field `ALTER field_name TYPE new_cql_datatype` - however, I think if you're doing this, you probably need to redeploy your application anyways since the types are changing.

PR #532, opened 2022-08-15
Reduce address comparisons for network topology replica calculation

mpenick:
This uses a `DenseHashSet` to keep prevent duplicate replicas instead of doing a linear scan through the existing replicas. I'm seeing around a 4.5x speed up for larger replication factors (rf = 54).

### DIFF
--- a/src/host.hpp
+++ b/src/host.hpp
@@ -304,4 +304,24 @@ bool remove_host(CopyOnWriteHostVec& hosts, const Address& address);
 
 }}} // namespace datastax::internal::core
 
+
+namespace std {
+
+#if defined(HASH_IN_TR1) && !defined(_WIN32)
+namespace tr1 {
+#endif
+
+template <>
+struct hash<datastax::internal::core::Host*> {
+  size_t operator()(const datastax::internal::core::Host* host) const {
+    return host->address().hash_code();
+  }
+};
+
+#if defined(HASH_IN_TR1) && !defined(_WIN32)
+} // namespace tr1
+#endif
+
+} // namespace std
+
 #endif
--- a/src/token_map_impl.hpp
+++ b/src/token_map_impl.hpp
@@ -388,6 +388,14 @@ inline bool add_replica(CopyOnWriteHostVec& hosts, const Host::Ptr& host) {
   return true;
 }
 
+class RawPtrHostSet : public DenseHashSet<Host*> {
+public:
+  RawPtrHostSet() {
+    set_empty_key(0x0);
+    set_deleted_key(reinterpret_cast<Host*>(0x1));
+  }
+};
+
 template <class Partitioner>
 void ReplicationStrategy<Partitioner>::build_replicas_network_topology(
     const TokenHostVec& tokens, const DatacenterMap& datacenters, TokenReplicasVec& result) const {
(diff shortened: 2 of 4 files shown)


tests/src/unit/tests/test_token_map.cpp
```diff
@@ -129,12 +129,12 @@ TEST(TokenMapUnitTest, Murmur3MultipleTokensPerHost) {
...
-  size_t num_racks = 3;
-  size_t num_hosts = 4;
+  size_t num_dcs = 2;
+  size_t num_racks = 1;
+  size_t num_hosts = 27;
   size_t num_vnodes = 256;
-  size_t replication_factor = 3;
-  size_t total_replicas = std::min(num_hosts, replication_factor) * num_dcs;
+  size_t replication_factor = 54;
+  size_t total_replicas = replication_factor;
```
mpenick, line 137: This should likely be reverted.

tests/src/unit/test_token_map_utils.hpp
```diff
@@ -79,6 +79,8 @@ class BufferBuilder {
 
   static size_t size_of(const String& value) { return value.size(); }
 
+  static size_t size_of(const Address& value) { char buf[16]; return value.to_inet(buf); }
```
mpenick, line 82: These changes fix test warnings.

src/token_map_impl.hpp
```diff
@@ -476,15 +487,17 @@ void ReplicationStrategy<Partitioner>::build_replicas_network_topology(
...
+          replicas->push_back(Host::Ptr(host));
           ++replica_count_this_dc;
         }
       } else {
         TokenHostQueue& skipped_endpoints_this_dc = dc_rack_info.skipped_endpoints;
         if (racks_observed_this_dc.count(rack) > 0) {
           skipped_endpoints_this_dc.push_back(curr_token_it);
         } else {
-          if (add_replica(replicas, Host::Ptr(host))) {
+          if (replicas_set.insert(host).second) {
```
zakalibit, line 499: why not just embed the changes in the `add_replica()`?
another option is to try keep `replicas` sorted and use binary search? i.e. use 
`std::sort()`, or even `std::stable_sort()` that should be faster for sorted or, almost sorted data, then just use `std::lower_bound()`  with custom `Comparator` to check if the host needs to be added.

src/host.hpp
```diff
@@ -304,4 +304,24 @@ bool remove_host(CopyOnWriteHostVec& hosts, const Address& address);
...
+namespace std {
+
+#if defined(HASH_IN_TR1) && !defined(_WIN32)
+namespace tr1 {
+#endif
+
+template <>
+struct hash<datastax::internal::core::Host*> {
+  size_t operator()(const datastax::internal::core::Host* host) const {
+    return host->address().hash_code();
```
mpenick, line 317: This conflates pointer identity with `Host`'s address identity and isn't correct. I wonder if pointer identity is enough.

src/token_map_impl.hpp
mpenick: The token order of `replicas` matters.

zakalibit: still could embed the logic in to `add_replica()`

PR #533, opened 2022-09-26
Adapt MemoryOrder definition for C++ 20

SeverinLeonhardt:
When compiling the code with C++ 20 the following error message is produced:

    enumerator value for ‘MEMORY_ORDER_RELAXED’ must have integral or unscoped enumeration type

This is because with C++ 20 the type of those values changed to an enum class. The `static_cast` extracts the numeric values so they can be used here as before.

### DIFF
--- a/src/atomic/atomic_std.hpp
+++ b/src/atomic/atomic_std.hpp
@@ -22,12 +22,12 @@
 namespace datastax { namespace internal {
 
 enum MemoryOrder {
-  MEMORY_ORDER_RELAXED = std::memory_order_relaxed,
-  MEMORY_ORDER_CONSUME = std::memory_order_consume,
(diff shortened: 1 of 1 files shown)


absurdfarce:
Thanks for the contributions @SeverinLeonhardt, this one as well as the other two PRs.  Unfortunately I've been completely swamped recently (thus the delay in getting back to you) but I'll try to take a look at these PRs next week!

absurdfarce:
Finally got a chance to take a look at this one @SeverinLeonhardt.  That's a nice and elegant way to solve that issue... and degrades nicely for the C++11 case (which is the default).

Many thanks for the solid contribution!  And again, my apologies for the delay; we're trying to get back on track here.

PR #534, opened 2022-09-26
Fix signed/unsigned mismatch

SeverinLeonhardt:
This fixes the following warnings from MSVC:

    warning C4245: 'initializing': conversion from 'int' to 'datastax::StringRef::size_type', signed/unsigned mismatch
    warning C4245: '=': conversion from 'int' to 'DWORD', signed/unsigned mismatch

### DIFF
--- a/src/string_ref.cpp
+++ b/src/string_ref.cpp
@@ -18,6 +18,6 @@
 
 namespace datastax {
 
-const StringRef::size_type StringRef::npos = -1;
+const StringRef::size_type StringRef::npos = std::numeric_limits<StringRef::size_type>::max();
 
 } // namespace datastax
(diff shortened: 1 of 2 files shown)


absurdfarce:
Thanks for the PR @SeverinLeonhardt!  These seem like pretty nice improvements; I had originally wondered about just casting the npos assignment but after thinking about it a bit I think your solution is the better option.

As discussed on the other PR I'm going to hold on merging this until we can get our build infrastructure issues sorted out.

absurdfarce:
Per comments on #535 merging this now so that it can be included in upcoming release.

Thanks (again) @SeverinLeonhardt !

PR #536, opened 2023-01-02
Add rack-aware load balancing policy

martin-sucha:
We need to prefer local rack as there are higher network costs
when communicating with nodes in remote rack.

This policy prefers nodes from the local rack, then local datacenter
and then other nodes.

The new RackAwarePolicy is similar to DCAwarePolicy,
but does not have the deprecated options.
TokenAwarePolicy and other code needed to be modified
so that the local rack is propagated.

The TokenAware policy was changed to prefer replicas in remote
rack / remote DC before trying non-replica nodes.
It does not make much sense to not try the replicas and
trying the replicas simplifies the code as now we have three
levels local/remote/remote2.

This change might not be backwards-compatible,
we don't know what exactly this project guarantees in terms of
backwards compatibility.

Co-Authored-By: Peter Navrátil <peter.navratil@kiwi.com>

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -2197,6 +2197,50 @@ cass_cluster_set_load_balance_dc_aware_n(CassCluster* cluster,
                                          cass_bool_t allow_remote_dcs_for_local_cl);
 
 /**
+ * Configures the cluster to use Rack-aware load balancing.
+ * For each query, all live nodes in a primary 'local' rack are tried first,
+ * followed by nodes from local DC and then nodes from other DCs.
+ *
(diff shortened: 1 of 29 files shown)


absurdfarce:
Thanks for the PR @martin-sucha!  I'll try to take a look at this as soon as I can but unfortunately I'm completely buried by other work at the moment so fair warning... it might take a while. :(

In the meantime... have you signed the Contributor License Agreement for contributions to DataStax open source projects? If not you can find it at https://cla.datastax.com/.

Thanks again!

PR #538, opened 2023-02-28
Prioritize ${OPENSSL_ROOT_DIR} over system paths

diku89:
This commit reorders the locations where CMake will look for openssl, starting with ${OPENSSL_ROOT_DIR} and then proceeding to system install locations.

This allows the user to provide a different Openssl version with -DOPENSSL_ROOT_DIR=<path/to/openssl>

### DIFF
--- a/cmake/FindOpenSSL.cmake
+++ b/cmake/FindOpenSSL.cmake
@@ -82,8 +82,8 @@ if (WIN32)
   unset(_programfiles)
 else ()
   set(_OPENSSL_ROOT_HINTS
-    /usr/local/opt/openssl
     ${OPENSSL_ROOT_DIR}
+    /usr/local/opt/openssl
     ENV OPENSSL_ROOT_DIR
     )
 endif ()


absurdfarce:
Greetings @diku89 and thanks for the contribution! Apologies for taking so long to get back to you.

Have you signed the Contributor License Agreement for contributions to DataStax open source projects? If not you can find it at https://cla.datastax.com/. Thanks!

diku89:
Hello. Thanks for the heads up. I just signed it.

absurdfarce:
Hey @diku89 , just so I have this right in my own head... the goal here is to make sure that an OpenSSL install supplied via the build param is used in preference over an existing install, correct?  That makes sense to me, so I think we're largely good if that's the case.  I need to test a few things locally but this seems pretty sane to me.

One thing: the syntax `ENV OPENSSL_ROOT_DIR` on line 87 seems to refer to specifying an OpenSSL install via an environment variable.  If we're preferring an OpenSSL install specified via a build param to a hard-coded value seems like we should prefer one specified via an environment variable as well.  So with that in mind it seems reasonable to make the hard-coded value(s) the _last_ thing in this list.

absurdfarce:
After some testing I can confirm that this is definitely not behaving as expected.  Here's what we have from the current behaviour in base, when passing props via the command-line and via env vars:

- Base using existing config:
  -  -- Found OpenSSL: /usr/local/opt/openssl/lib/libssl.so (found version "1.1.1q") 
- LIBUV_ROOT_DIR=/work/local/libuv-1.44.2/ cmake -DOPENSSL_ROOT_DIR=/work/local/openssl ..
  - -- Found OpenSSL: /usr/local/opt/openssl/lib/libssl.so (found version "1.1.1q")
- LIBUV_ROOT_DIR=/work/local/libuv-1.44.2/ OPENSSL_ROOT_DIR=/work/local/openssl cmake ..
  - -- Found OpenSSL: /usr/local/opt/openssl/lib/libssl.so (found version "1.1.1q")

Just moving the default /usr/local/opt/openssl config down seems to leave us in a much better spot:

- Base using existing config:
  -  -- Found OpenSSL: /usr/local/opt/openssl/lib/libssl.so (found version "1.1.1q")
- LIBUV_ROOT_DIR=/work/local/libuv-1.44.2/ cmake -DOPENSSL_ROOT_DIR=/work/local/openssl ..
  - -- Found OpenSSL: /work/local/openssl/lib/libssl.so (found version "1.1.1q")
- LIBUV_ROOT_DIR=/work/local/libuv-1.44.2/ OPENSSL_ROOT_DIR=/work/local/openssl cmake ..
  - -- Found OpenSSL: /work/local/openssl/lib/libssl.so (found version "1.1.1q")

I'm going to merge this PR as it stands and tweak it slightly after the fact so that both the CLI prop and the env var can override the default.

Huge thanks to @diku89 for the PR... and for your patience!

absurdfarce:
Quick note for the record: follow up commit mentioned above is [here](https://github.com/datastax/cpp-driver/commit/1e8c520e611c7c64beb61ac01fbb9e660599e335).

PR #540, opened 2023-05-19
fix the missing OS_DISTRO env variable in some steps

weideng1:


joao-r-reis (commented):
I would have preferred to use a local variable instead of an environment variable because I'm not sure if jenkins will try to make the scope of that variable shared across different runners and then we could get into a race condition if both `post` sections run at the same time on different runners.

I saw this happen in Node.js builds because we are using an environment variable to know whether a runner already notified slack that the build has started and some times we get a duplicate "build started" message on slack because of this: https://github.com/datastax/nodejs-driver/blob/d29e63156f7ca6e4597ad9a23302e47770408d9a/Jenkinsfile#L392C32-L394

joao-r-reis (approved):

PR #543, opened 2023-07-25
CPP-971 Fix for beta protocol support in integration tests

absurdfarce:


tests/src/integration/integration.cpp
```diff
@@ -373,12 +373,7 @@ test::driver::Cluster Integration::default_cluster(bool is_with_default_contact_
   if (is_with_default_contact_points) {
     cluster.with_contact_points(contact_points_);
   }
-  if (Options::is_cassandra() && server_version_ >= "4.0" &&
-      protocol_version_ == CASS_PROTOCOL_VERSION_V4) {
-    cluster.with_beta_protocol(is_beta_protocol_);
-  } else {
-    cluster.with_protocol_version(protocol_version_);
-  }
```
absurdfarce, line 381: Cassandra 4.0 supports v5 as a non-beta protocol so the block above doesn't really make sense anymore

tests/src/integration/options.cpp
```diff
@@ -48,7 +48,7 @@ std::string Options::public_key_ = "public.key";
 std::string Options::private_key_ = "private.key";
 bool Options::is_verbose_ccm_ = false;
 bool Options::is_verbose_integration_ = false;
-bool Options::is_beta_protocol_ = true;
+bool Options::is_beta_protocol_ = false;
```
absurdfarce, line 51: Not strictly necessary (the fix above is more than adequate to address the problem) but more correct given the current state of things.  The driver will not recognize v5-beta in other integration tests but since v4 is the highest protocol version we officially support this _shouldn't_ be a major obstacle.

weideng1 (approved):

tests/src/integration/tests/test_set_keyspace.cpp
```diff
@@ -175,7 +175,7 @@ CASSANDRA_INTEGRATION_TEST_F(SetKeyspaceTests, QueryNotSupported) {
  */
 CASSANDRA_INTEGRATION_TEST_F(SetKeyspaceTests, QueryWithNoSessionKeyspace) {
   CHECK_FAILURE;
-  CHECK_VERSION(4.0.0);
+  CHECK_PROTOCOL_VERSION(CASS_PROTOCOL_VERSION_V5);
```
absurdfarce, line 178: The changes above mean that we're now _not_ trying to connect as a beta version of v5.  This means that for C* 4.0 we'll now connect with v4 (the max version actually supported by this driver).  This change is actually a problem for some tests (like this one) which rely on v5 features [1] but can't be executed against an actual server since we only now have v5 support.

To be more explicit about these requirements we constrain these tests by protocol version used for a given connection rather than by a Cassandra version.

[1] Set keyspaces for statements was added in protocol v5

absurdfarce:
Confirmed that this PR gets us working with C* 4.0 again.  It also addresses a number of new test failures which sprang up once we removed the beta-v5 support and shifted everything down to v4.

PR #546, opened 2023-07-26
CPP-973: fix column count assert in the test

weideng1:


absurdfarce (approved):
Testing confirms this change resolves the issue as expected.

PR #547, opened 2023-07-27
CPP-975: remove special handling for 4.0 in ExecutionProfileTest.Inte…

weideng1:
…gration_Cassandra_Consistency as the error message is consistent with earlier C* versions


absurdfarce (approved):
Testing confirms this change resolves the issue as expected.

PR #548, opened 2023-07-27
CPP-972: make expected_message more generic so it can match C* 4.0 re…

weideng1:
…sult.error_message in test ServerSideFailureTests.Integration_Cassandra_ErrorFunctionAlreadyExists

### DIFF
--- a/tests/src/integration/tests/test_server_side_failure.cpp
+++ b/tests/src/integration/tests/test_server_side_failure.cpp
@@ -158,8 +158,7 @@ CASSANDRA_INTEGRATION_TEST_F(ServerSideFailureTests, ErrorFunctionAlreadyExists)
   session_.execute(create_function_query);
   Result result = session_.execute(Statement(create_function_query), false);
   EXPECT_EQ(CASS_ERROR_SERVER_INVALID_QUERY, result.error_code());
-  EXPECT_TRUE(result.error_message().find("(double) -> double already exists") !=
-              std::string::npos);
+  EXPECT_TRUE(result.error_message().find(" already exists") != std::string::npos);
 
   ErrorResult error_result = result.error_result();
   ASSERT_TRUE(error_result);


tests/src/integration/tests/test_server_side_failure.cpp
```diff
@@ -158,8 +158,7 @@ CASSANDRA_INTEGRATION_TEST_F(ServerSideFailureTests, ErrorFunctionAlreadyExists)
   session_.execute(create_function_query);
   Result result = session_.execute(Statement(create_function_query), false);
   EXPECT_EQ(CASS_ERROR_SERVER_INVALID_QUERY, result.error_code());
-  EXPECT_TRUE(result.error_message().find("(double) -> double already exists") !=
-              std::string::npos);
+  EXPECT_TRUE(result.error_message().find(" already exists") != std::string::npos);
```
absurdfarce, line 161: I feel like we might be losing specificity here when we don't necessarily have to:

```
[cqlsh 6.0.0 | Cassandra 4.0.2 | CQL spec 3.4.5 | Native protocol v5]
Use HELP for help.
cqlsh> use ks;
cqlsh:ks> CREATE FUNCTION already_exists_function(value double) RETURNS NULL ON NULL INPUT RETURNS double LANGUAGE java AS 'return 3.14;' ;
cqlsh:ks> CREATE FUNCTION already_exists_function(value double) RETURNS NULL ON NULL INPUT RETURNS double LANGUAGE java AS 'return 3.14;' ;
InvalidRequest: Error from server: code=2200 [Invalid query] message="Function 'already_exists_function' already exists"
```

```
[cqlsh 5.0.1 | Cassandra 3.11.15 | CQL spec 3.4.4 | Native protocol v4]
Use HELP for help.
cqlsh> use testy;
cqlsh:testy> CREATE FUNCTION already_exists_function(value double) RETURNS NULL ON NULL INPUT RETURNS double LANGUAGE java AS 'return 3.14;' ;
cqlsh:testy> CREATE FUNCTION already_exists_function(value double) RETURNS NULL ON NULL INPUT RETURNS double LANGUAGE java AS 'return 3.14;' ;
InvalidRequest: Error from server: code=2200 [Invalid query] message="Function testy.already_exists_function : (double) -> double already exists"
```

Seems like what we really want to check for here is something like "Function(.*)already_exists_function(.*) already exists".  We should be able to do that via something like:

* Get the position of "Function" as pos1
* EXPECT_TRUE(pos1 != npos)
* Get the position of the function name as pos2
* EXPECT_TRUE(pos2 != npos)
* EXPECT_TRUE(pos2 > pos1) (since it should always come after it in the string)
* Get the position of " already exists" as pos3
* EXPECT_TRUE(pos3 != npos)
* EXPECT_TRUE(pos3 > pos2) (as above)

absurdfarce (changes requested):
I think we can make this test quite a bit more granular and still keep it passing across all tested Cassandra versions.

PR #550, opened 2023-07-28
CPP-974 PreparedMetadataTests.Integration_Cassandra_AlterProperlyUpdatesColumnCount fails with C* 4.0+

absurdfarce:


absurdfarce:
This fix has been folded into the much more expansive set of fixes implemented for [CPP-966](https://datastax-oss.atlassian.net/browse/CPP-966) (relevant PR [here](https://github.com/datastax/cpp-driver/pull/551))

PR #551, opened 2023-07-28
CPP-966 Compilation warnings after cass_set_use_hostname_resolution() marked as deprecated

absurdfarce:
Along with some other miscellaneous fixes as well:

* Remove beta protocol support from the CLI options to the integration test as well as the test cluster itself
* Clean up some error messages

### DIFF
--- a/tests/src/integration/ccm/bridge.cpp
+++ b/tests/src/integration/ccm/bridge.cpp
@@ -1491,7 +1491,7 @@ CCM::Bridge::generate_create_updateconf_command(CassVersion cassandra_version) {
     updateconf_command.push_back("enable_scripted_user_defined_functions:true");
   }
 
-  if (cassandra_version >= "4.0.0") {
+  if (cassandra_version >= "4.0.0" && !is_dse()) {
     updateconf_command.push_back("enable_materialized_views:true");
     updateconf_command.push_back("enable_user_defined_functions:true");
   }
--- a/tests/src/integration/objects/cluster.hpp
+++ b/tests/src/integration/objects/cluster.hpp
@@ -207,22 +207,6 @@ public:
   }
 
   /**
-   * Enable/Disable the use of hostname resolution
-   *
-   * This is useful for authentication (Kerberos) or encryption (SSL)
-   * services that require a valid hostname for verification.
-   *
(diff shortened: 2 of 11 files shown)


tests/src/integration/ccm/bridge.cpp
```diff
@@ -1491,7 +1491,7 @@ CCM::Bridge::generate_create_updateconf_command(CassVersion cassandra_version) {
     updateconf_command.push_back("enable_scripted_user_defined_functions:true");
   }
 
-  if (cassandra_version >= "4.0.0") {
+  if (cassandra_version >= "4.0.0" && !is_dse()) {
```
absurdfarce, line 1494: A small fix to get DSE working for integration tests.  IIRC @weideng1 has something similar to this in another PR.

tests/src/integration/objects/cluster.hpp
```diff
@@ -206,22 +193,6 @@ class Cluster : public Object<CassCluster, cass_cluster_free> {
...
-   * @param enable True if hostname resolution should be enabled; false
-   *               otherwise (default: true)
-   * @return Cluster object
-   */
-  Cluster& with_hostname_resolution(bool enable = true) {
-    EXPECT_EQ(CASS_OK, cass_cluster_set_use_hostname_resolution(
-                           get(), (enable == true ? cass_true : cass_false)));
-    return *this;
-  }
-
```
absurdfarce, line 224: The heart of the matter.  The use of this function was the immediate cause of the slew of deprecated warnings when building the integration tests.

tests/src/integration/tests/test_dse_auth.cpp
```diff
@@ -203,7 +202,6 @@ class AuthenticationTest : public DseIntegration {
     Cluster cluster = dse::Cluster::build()
                           .with_plaintext_authenticator(username, password)
                           .with_contact_points(contact_points_)
-                          .with_hostname_resolution(true)
```
absurdfarce, line 206: These two usages are the only places where with_hostname_resolution() was actually used.  Manually confirmed this test passes for all supported versions of DSE after this change.

weideng1 (approved):

PR #552, opened 2023-07-31
2 17 0 doc updates

absurdfarce:


README.md
```diff
@@ -209,17 +208,16 @@ specific language governing permissions and limitations under the License.
...
 [Getting Started]: http://docs.datastax.com/en/developer/cpp-driver/latest/topics
 [Building]: http://docs.datastax.com/en/developer/cpp-driver/latest/topics/building
-[centos-6-dependencies]: http://downloads.datastax.com/cpp-driver/centos/6/dependencies
 [centos-7-dependencies]: http://downloads.datastax.com/cpp-driver/centos/7/dependencies
-[centos-8-dependencies]: http://downloads.datastax.com/cpp-driver/centos/8/dependencies
-[ubuntu-14-04-dependencies]: http://downloads.datastax.com/cpp-driver/ubuntu/14.04/dependencies
-[ubuntu-16-04-dependencies]: http://downloads.datastax.com/cpp-driver/ubuntu/16.04/dependencies
-[ubuntu-18-04-dependencies]: http://downloads.datastax.com/cpp-driver/ubuntu/18.04/dependencies
+[rocky-linux-8-dependencies]: http://downloads.datastax.com/cpp-driver/rocky/8/dependencies
+[rocky-linux-9-dependencies]: http://downloads.datastax.com/cpp-driver/rocky/9/dependencies
```
absurdfarce, line 218: These should be added as part of the release process

README.md
```diff
@@ -209,17 +208,16 @@ specific language governing permissions and limitations under the License.
 [Examples]: examples/
 [download server]: http://downloads.datastax.com/cpp-driver/
 [GitHub]: https://github.com/datastax/cpp-driver
-[cpp-driver-compatability-matrix]: https://docs.datastax.com/en/developer/driver-matrix/doc/cppDrivers.html#cpp-drivers
+[cpp-driver-compatability-matrix]: https://docs.datastax.com/en/driver-matrix/docs/cpp-drivers.html
```
absurdfarce, line 211: Link which needed to be updated in light of the site rework

README.md
```diff
@@ -19,12 +19,11 @@ provided under the `dependencies` directory for each platform (if applicable).
...
 distribution:
 
-* [CentOS 6][centos-6-dependencies]
 * [CentOS 7][centos-7-dependencies]
-* [CentOS 8][centos-8-dependencies]
-* [Ubuntu 14.04][ubuntu-14-04-dependencies]
-* [Ubuntu 16.04][ubuntu-16-04-dependencies]
-* [Ubuntu 18.04][ubuntu-18-04-dependencies]
+* [Rocky Linux 8][rocky-linux-8-dependencies]
+* [Rocky Linux 9][rocky-linux-9-dependencies]
```
absurdfarce, line 24: Might want to consider whether we want major.minor for Rocky releases or just major here.

README.md
```diff
@@ -87,9 +86,9 @@ dependencies for a specific platform.
 This driver works exclusively with the Cassandra Query Language v3 (CQL3) and
 Cassandra's native protocol. The current version works with:
 
-* Apache Cassandra® versions 2.1, 2.2 and 3.0+
+* Apache Cassandra® versions 3.0.x, 3.11.x and 4.0.x or DSE versions 6.8.x and 5.1.x
 * Architectures: 32-bit (x86) and 64-bit (x64)
-* Compilers: GCC 4.1.2+, Clang 3.4+, and MSVC 2010/2012/2013/2015/2017/2019
+* Compilers: GCC 4.8.5+, Clang 3.4+, and MSVC 2013/2015/2017/2019
```
absurdfarce, line 91: Update here based on lowest gcc version implemented by our various testing platforms.  Taking the oldest version we still support from each group we have the following:

```
Rocky 8 = g++ (GCC) 8.5.0 20210514 (Red Hat 8.5.0-18)
Ubuntu Focal = g++ (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
CentOS 7 = g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)
```

This leads to the use of 4.8.5, but I wonder if this isn't kind of misleading.

I'm wondering if it doesn't make more sense to change this section to indicate what versions we're testing against in more detail rather than saying we support version X up.  I'm thinking of listing all the supported OS versions + their corresponding gcc version.  On the Windows side we'd just keep consistent with the Visual Studio versions we use for the build.

absurdfarce: This has been re-worked in subsequent commits

README.md
```diff
@@ -10,22 +10,28 @@ products:
...
-__Note__: CentOS and Ubuntu use the version of OpenSSL and zlib provided with the
-distribution:
-
-* [CentOS 6][centos-6-dependencies]
-* [CentOS 7][centos-7-dependencies]
-* [CentOS 8][centos-8-dependencies]
-* [Ubuntu 14.04][ubuntu-14-04-dependencies]
-* [Ubuntu 16.04][ubuntu-16-04-dependencies]
-* [Ubuntu 18.04][ubuntu-18-04-dependencies]
-* [Windows][windows-dependencies]
```
absurdfarce, line 28: Re-worked this section (and removed the links) since there is no common location in Artifactory that can be used as an entry point for dependencies.  We can point users at the "builds" directory but underneath there we have subdirectories for version number and build number.. so it seemed better to provide guidance on how to find the deps rather than try to come up with some arbitrary URL scheme to point at them.

topics/installation/README.md
```diff
@@ -2,66 +2,17 @@
 
 ## Packages
 
-Pre-built packages are available for CentOS 6/7/8, Ubuntu 14.04/16.04/18.04 and
-Windows 7 SP1 and above.
+Pre-built packages are available for CentOS 7, Ubuntu 20.04/22.04 and
+Rocky Linux 8 and 9.  All packages are available from our
```
weideng1, line 6: @absurdfarce Did we miss mentioning Windows packages here?

absurdfarce: Yes, will add.

PR #556, opened 2023-09-19
CPP-991 Add extraction of version string for OpenSSL 3.x

absurdfarce:


absurdfarce:
Change has been correctly printing version strings for OpenSSL 3.x as well as 1.1.x on multiple builds now.  Calling this good.

PR #558, opened 2023-09-22
CPP-988 Fix race condition in monotonic_timestamp()

absurdfarce:


src/uuids.cpp
```diff
@@ -269,7 +269,7 @@ uint64_t UuidGen::monotonic_timestamp() {
     } else {
       uint64_t last_ms = to_milliseconds(last);
       if (to_milliseconds(now) < last_ms) {
-        return last_timestamp_.fetch_add(1);
+        return (last_timestamp_.fetch_add(1) + 1);
```
absurdfarce, line 272: In review @hhughes pointed out that both the now < last_ms and now == last_ms cases effectively reduce to "return 1 + the last timestamp".  This can be implemented more succinctly via a single CAS op along the lines of the following:

```
uint64_t UuidGen::monotonic_timestamp() {
  while (true) {
    uint64_t now = from_unix_timestamp(get_time_since_epoch_ms());
    uint64_t last = last_timestamp_.load();
    if (now > last) {
      if (last_timestamp_.compare_exchange_strong(last, now)) {
        return now;
      }
    } else {
      uint64_t candidate = last + 1;
      if (last_timestamp_.compare_exchange_strong(last, candidate)) {
        return candidate;
      }
    }
  }
}
```

In testing this seems to perform roughly as well as my initial change (i.e. no duplicate UUIDs created in multiple runs of the sample client) and it's significantly simpler to grok.  I'm inclined to go with it unless somebody can cite a reason not to.

absurdfarce: Moving forward with this change as an intermediate commit.  In code review @hhughes suggested an additional optimization which should make the code even more readable; I'll try that next.

absurdfarce:
After thinking about the testing question for a while I'm not sure there is a clear answer that gives us obvious benefit at this point.  I've filed [CPP-992](https://datastax-oss.atlassian.net/browse/CPP-992) to consider this more fully; I feel like we have to give this some serious thought.

With that in mind I believe this ticket is good to merge as it stands.

PR #561, opened 2023-10-06
CPP-964 Add refresh-interval support for histogram metrics

absurdfarce:
Port of Java functionality as defined [here](https://github.com/datastax/java-driver/blob/4.x/core/src/main/resources/reference.conf#L1554-L1571).


absurdfarce:
Some sample code for playing with the refresh rate in a real app:

```
#include "cassandra.h"
/* Use "#include <dse.h>" when connecting to DataStax Enterpise */
#include <stdio.h>
#include <unistd.h>

void on_log(const CassLogMessage* message, void* data) {
  FILE* log_file = (FILE*)data;
  fprintf(log_file, "%u.%03u [%s] (%s:%d:%s): %s\n", (unsigned int)(message->time_ms / 1000),
          (unsigned int)(message->time_ms % 1000), cass_log_level_string(message->severity),
          message->file, message->line, message->function, message->message);
}

void execute_query(CassStatement* statement, CassSession* session) {

    CassFuture* result_future = cass_session_execute(session, statement);

    if (cass_future_error_code(result_future) == CASS_OK) {
      /* Retrieve result set and get the first row */
      const CassResult* result = cass_future_get_result(result_future);
      const CassRow* row = cass_result_first_row(result);

      if (row) {
        const CassValue* value = cass_row_get_column_by_name(row, "release_version");

        const char* release_version;
        size_t release_version_length;
        cass_value_get_string(value, &release_version, &release_version_length);
        printf("release_version: '%.*s'\n", (int)release_version_length, release_version);
      }

      cass_result_free(result);
    } else {
      /* Handle error */
      const char* message;
      size_t message_length;
      cass_future_error_message(result_future, &message, &message_length);
      fprintf(stderr, "Unable to run query: '%.*s'\n", (int)message_length, message);
    }

    cass_future_free(result_future);
}

int main(int argc, char* argv[]) {

  FILE* log_file = fopen("driver.log", "w+");
  if (log_file == NULL) {
    fprintf(stderr, "Unable to open log file\n");
  }

  /* Log configuration *MUST* be done before any other driver call */
  cass_log_set_level(CASS_LOG_INFO);
  cass_log_set_callback(on_log, (void*)log_file);

  /* Setup and connect to cluster */
  CassFuture* connect_future = NULL;
  CassCluster* cluster = cass_cluster_new();
  CassSession* session = cass_session_new();
  const char* hosts = "127.0.0.1";
  if (argc > 1) {
    hosts = argv[1];
  }
  
  /* Add contact points */
  cass_cluster_set_contact_points(cluster, hosts);

  /* Specify a histogram refresh period */
  cass_cluster_set_histogram_refresh_interval(cluster, 5000);

  /* Provide the cluster object as configuration to connect the session */
  connect_future = cass_session_connect(session, cluster);

  if (cass_future_error_code(connect_future) == CASS_OK) {
    CassFuture* close_future = NULL;

    /* Build statement and execute query */
    const char* query = "SELECT release_version FROM system.local";
    CassStatement* statement = cass_statement_new(query, 0);
    CassMetrics metrics;

    for (int i = 1; i < 10; ++i) {

      execute_query(statement, session);
      sleep(1);

      cass_session_get_metrics(session, &metrics);
      printf("Requests min: %u\n", metrics.requests.min);
    }

    /* Close the session */
    cass_statement_free(statement);
    close_future = cass_session_close(session);
    cass_future_wait(close_future);
    cass_future_free(close_future);
  } else {
    /* Handle error */
    const char* message;
    size_t message_length;
    cass_future_error_message(connect_future, &message, &message_length);
    fprintf(stderr, "Unable to connect: '%.*s'\n", (int)message_length, message);
  }

  cass_future_free(connect_future);
  cass_cluster_free(cluster);
  cass_session_free(session);

  return 0;
}
```

src/constants.hpp
```diff
@@ -142,6 +142,7 @@
 #define CASS_DEFAULT_MAX_TRACING_DATA_WAIT_TIME_MS 15
 #define CASS_DEFAULT_RETRY_TRACING_DATA_WAIT_TIME_MS 3
 #define CASS_DEFAULT_TRACING_CONSISTENCY CASS_CONSISTENCY_ONE
+#define CASS_DEFAULT_HISTOGRAM_REFRESH_INTERVAL_NO_REFRESH 0
```
absurdfarce, line 145: I don't _love_ this name for the constant but... I couldn't come up with something I liked better.

src/metrics.hpp
```diff
@@ -268,9 +269,15 @@ class Metrics : public Allocated {
       int64_t percentile_999th;
     };
 
-    Histogram(ThreadState* thread_state)
+    Histogram(ThreadState* thread_state, unsigned refresh_interval = CASS_DEFAULT_HISTOGRAM_REFRESH_INTERVAL_NO_REFRESH)
         : thread_state_(thread_state)
-        , histograms_(new PerThreadHistogram[thread_state->max_threads()]) {
+        , histograms_(new PerThreadHistogram[thread_state->max_threads()])
+        , zero_snapshot_(Snapshot {0,0,0,0,0,0,0,0,0,0}) {
```
absurdfarce, line 275: This feels like something that should be a static value on the Histogram class but repeated attempts to setup such a value hit various issues that my knowledge of C++ wasn't subtle enough to address.  Since we're only creating one of these for a given metrics impl I wasn't _too_ worried about it (the size of a Snapshot shouldn't be a driving concern) but I concede it's not ideal.

src/metrics.hpp
```diff
@@ -286,38 +293,73 @@ class Metrics : public Allocated {
...
+
+        for (size_t i = 0; i < thread_state_->max_threads(); ++i) {
+          histograms_[i].add(histogram_);
+        }
+
+        cached_snapshot_ = build_new_snapshot(histogram_);
+        refresh_timestamp_ = now;
       }
+
+      copy_snapshot(cached_snapshot_, snapshot);
```
absurdfarce, line 330: Model here very much is the [impl of similar functionality](https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/metrics/HdrReservoir.java#L108-L142) in the 4.x Java driver

src/cluster_config.cpp
```diff
@@ -577,6 +577,14 @@ void cass_cluster_set_monitor_reporting_interval(CassCluster* cluster, unsigned
   cluster->config().set_monitor_reporting_interval_secs(interval_secs);
 }
 
+CassError cass_cluster_set_histogram_refresh_interval(CassCluster* cluster, unsigned refresh_interval) {
+  if (refresh_interval <= 0) {
```
hhughes, line 581: note: since the parameter is unsigned, we're never going to have a negative interval, and if this is called with -1 they will get a huge interval (arguably this is their fault for not reading the signature)

src/metrics.hpp
```diff
@@ -286,38 +293,89 @@ class Metrics : public Allocated {
...
+      uint64_t now = get_time_since_epoch_ms();
+      if (now - refresh_timestamp_ >= refresh_interval_) {
+
+        hdr_reset(histogram_);
+
+        for (size_t i = 0; i < thread_state_->max_threads(); ++i) {
+          histograms_[i].add(histogram_);
+        }
+
+        cached_snapshot_ = build_new_snapshot(histogram_);
```
hhughes, line 326: can we use `build_and_copy_snapshot(histogram_, cached_snapshot_)` here instead?

src/metrics.hpp
```diff
@@ -286,38 +293,89 @@ class Metrics : public Allocated {
...
-        snapshot->percentile_98th = hdr_value_at_percentile(h, 98.0);
-        snapshot->percentile_99th = hdr_value_at_percentile(h, 99.0);
-        snapshot->percentile_999th = hdr_value_at_percentile(h, 99.9);
+      // Refresh interval is in use.  If we've exceeded the interval clear histogram_,
+      // compute a new aggregate histogram and build (and cache) a new snapshot.  Otherwise
+      // just return the cached version.
+      uint64_t now = get_time_since_epoch_ms();
+      if (now - refresh_timestamp_ >= refresh_interval_) {
+
+        hdr_reset(histogram_);
```
hhughes, line 320: Something I found a bit funny is that if I expand `HistogramWithRefreshInterval` test to wait for the next snapshot but without any requests, i.e. adding to the end:
```
  test::Utils::msleep(1.2 * refresh_interval);
  histogram.get_snapshot(&snapshot);
  EXPECT_EQ(snapshot.min, 0);
  EXPECT_EQ(snapshot.max, 0);
  EXPECT_EQ(snapshot.median, 0);
  EXPECT_EQ(snapshot.percentile_75th, 0);
  EXPECT_EQ(snapshot.percentile_95th, 0);
  EXPECT_EQ(snapshot.percentile_98th, 0);
  EXPECT_EQ(snapshot.percentile_99th, 0);
  EXPECT_EQ(snapshot.percentile_999th, 0);
  EXPECT_EQ(snapshot.mean, 0);
  EXPECT_EQ(snapshot.stddev, 0);
```
This (unexpectedly?) fails because snapshot.min has value `INT64_MAX` - which comes from `hdr_reset`.  When this windowing change is not enabled the initial value of `min` is zero so I think having this value reported differently in a window when there are no requests will cause confusion for devs and possibly issues for downstream metricking systems. This seems like something that should be addressed before merging this change but happy to discuss.

src/metrics.hpp
hhughes, line 326: Nit: can we use `build_and_copy_snapshot(histogram_, &cached_snapshot_)` here to save an object creation?

hhughes (changes requested):

src/metrics.hpp
absurdfarce: Not obviously.  The whole point of build_and_copy_snapshot() was to avoid the creation of an intermediate Snapshot object and just update the "rv" Snapshot* pointer directly.  In this line we very explicitly want _to create_ that intermediate Snapshot object because then we're going to cache it.

src/metrics.hpp
```diff
@@ -286,38 +293,89 @@ class Metrics : public Allocated {
...
+        hdr_value_at_percentile(h, 95.0),
+        hdr_value_at_percentile(h, 98.0),
+        hdr_value_at_percentile(h, 99.0),
+        hdr_value_at_percentile(h, 99.9)
+      };
+    }
+
+    // Optimized version of chained build_new_snapshot -> copy_snapshot calls
+    // to avoid creation of an unnecessary intermediate struct
+    void build_and_copy_snapshot(hdr_histogram* h, Snapshot* to) const {
```
joao-r-reis, line 365: I'd probably remove the `build` word from this, it's not actually creating a new object like `build_new_snapshot` is.

Maybe `copy_snapshot_to_hdr_hist`?

joao-r-reis (commented):
Left a NIT comment, I don't have time to set up an env to build this locally and I see that Henry is already doing that so I'll leave that part to him for now 😅

src/metrics.hpp
absurdfarce: I went with the simpler "histogram_to_snapshot"... hopefully that's clearer.

src/metrics.hpp
absurdfarce: After looking at this I'm kinda wondering if we've been doing this wrong the whole time.

We're populating the snapshot with the results of hdr_min() which looks like this:

```
int64_t hdr_min(struct hdr_histogram* h)
{
    if (0 < hdr_count_at_index(h, 0))
    {
        return 0;
    }

    return non_zero_min(h);
}
```

Near as I can tell this function has no notion of "the histogram is empty", which causes it to always enter the "return non-zero" case.  The short-circuit return above covers the case in which we've actually recorded one or more values of zero in the histogram.  In that case we return zero, but in all other cases we return the initial value of INT64_MAX.  Upshot is a clear distinction between "we saw no values" and "we saw values and the smallest one was zero".

The snapshot struct munges these two cases together by returning zeroes in the empty case, but it seems far more disruptive to try and change that now vs. making the new refresh case work the same as the existing interface.  To that end we'll just return the zero Snapshot if we have an empty histogram.

absurdfarce:
Hey @hhughes and/or @joao-r-reis ; you guys good after the last round of changes?  I think everything has been addressed but wanted a final confirmation from one or (ideally) both of you on that point.

joao-r-reis:
Yeah the issue I brought up is resolved.

hhughes (approved):
LGTM

PR #562, opened 2023-10-10
On macOS search for libuv in /opt/homebrew, set CMAKE_MACOSX_RPATH to fix warning.

hhughes:


cmake/FindLibuv.cmake
```diff
@@ -25,8 +25,13 @@ if(NOT WIN32)
   set(_LIBUV_ROOT_PATHS "/usr/"
                         "/usr/local/")
   if(_LIBUV_FOUND)
-    set(_LIBUV_ROOT_PATHS ${_LIBUV_ROOT_PATHS}
-                          ${_LIBUV_LIBDIR})
+    if(APPLE)
+      set(_LIBUV_ROOT_PATHS ${_LIBUV_ROOT_PATHS}
+              ${_LIBUV_PREFIX})
```
hhughes, line 30: on macos uv.h ends up in `/opt/homebrew/Cellar/libuv/1.46.0/include/uv.h`, but `_LIBUV_LIBDIR` resolves to `/opt/homebrew/Cellar/libuv/1.46.0/lib`

CMakeLists.txt
```diff
@@ -170,8 +170,11 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
 
   # OpenSSL is deprecated on later versions of Mac OS X. The long-term solution
   # is to provide a CommonCryto implementation.
-  if (APPLE AND CASS_USE_OPENSSL)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated-declarations")
+  if (APPLE)
+    set(CMAKE_MACOSX_RPATH 1)
```
hhughes, line 174: Without this `cmake` logs warning:
```
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   cassandra

This warning is for project developers.  Use -Wno-dev to suppress it.
```

cmake/FindLibuv.cmake
absurdfarce: You should be able to address this by specifying the libuv root dir directly.  Try the following to see if it works:

```
LIBUV_ROOT_DIR=/opt/homebrew/Cellar/libuv/1.46.0 cmake ..
```

CMakeLists.txt
absurdfarce: Looks like this was added in the cmake version (2.8.12) we moved to as a minimum with the last release.  I'm hardly an expert on builds on MacOS but this seems pretty reasonable to me based on what I've read.

cmake/FindLibuv.cmake
hhughes: Thanks, that worked. I've changed this patch to update the building documentation to reflect that.

hhughes: Another possible solve is to have it search in `/opt/homebrew` when building on macOS which means it compiles without having to set additional env vars and keeps the search code common across all architectures (since we're just adding an extra common location). If this is still sketchy I can revert that and leave it as a comment in the docs.

absurdfarce: Oh, nice!  So adding /opt/homebrew to _LIBUV_ROOT_PATHS in the Apple case does make the build work without anything else?  I don't have a problem with that change at all; we actually do [something very similar to this](https://github.com/datastax/cpp-driver/blob/master/cmake/FindOpenSSL.cmake#L87) on the OpenSSL side.  /usr/local/opt/openssl isn't obviously an OSX path but it's not something you see on your average Linux system... and it does appear to be the spot where the Macs install OSX.

I don't think adding /opt/homebrew in the OSX case is janky at all.  If this fixes enables libuv detection automatically I'd say let's go with it.

absurdfarce: Confirmed directly with @hhughes that everything builds correctly on OSX with just the _LIBUV_ROOT_PATHS change above (as in there's no additional steps required to handle the libuv discovery/include bits).  Calling that good.

absurdfarce (approved):

PR #565, opened 2024-07-24
Remove centos7 and add Cassandra 5.0-beta1

joao-r-reis:
Removed CentOS7 because it's EOL.

Currently tests are failing and I think most (if not all) are related to cassandra.yml setting renaming that happened in C* 4.1. This is happening with other drivers as well. The Java driver already handled this in https://github.com/apache/cassandra-java-driver/pull/1924

C* 4.1 renamed some config settings but C* is still looking at the old names because of backwards compability reasons. However, C* fails to launch if it finds both the NEW name and the OLD name for a specific setting. The driver tests call ccm to add the OLD config setting but because the name is different from the one in cassadra.yml, the setting is added to cassandra.yml instead of replacing it.

I've already updated the AMI so it's already running the updated ccm that supports 5.0.

Note that you need to manually trigger the integration tests on jenkins, they don't run by default on the per-commit build.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -406,9 +406,9 @@ pipeline {
     choice(
       name: 'ADHOC_BUILD_AND_EXECUTE_TESTS_SERVER_VERSION',
       choices: [
-                '3.0',      // Previous Apache Cassandra
                 '3.11',     // Current Apache Cassandra
-                '4.0',      // Development Apache Cassandra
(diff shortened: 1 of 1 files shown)


absurdfarce:
This has been replaced by a combination of CPP-1001 (already merged) for the CentOS 7 removal and CPP-1002 for the Cassandra 5.0.x (and 4.1.x) support.

PR #566, opened 2024-10-22
CPP-1001 Remove CentOS 7 as a tested platform

absurdfarce:
And also remove references to Ubuntu Bionic while we're in here.


absurdfarce:
Thanks @lukasz-antoniak.  I'm calling this good for now.

PR #567, opened 2024-11-06
Initial version of a Windows config for Github Actions

absurdfarce:


.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
...
+    env:
+      CASS_DRIVER_LIBEV_INCLUDES: C:/vcpkg/packages/libev_x64-windows/include/libev
+      CASS_DRIVER_LIBEV_LIBS: C:/vcpkg/packages/libev_x64-windows/lib
+      LIBUV_ROOT_DIR: C:/vcpkg/packages/libuv_x64-windows
+      ZLIB_ROOT_DIR: C:/vcpkg/packages/zlib_x64-windows-static
+      KERBEROS_ROOT_DIR: C:/vcpkg/packages/krb5_x64-windows
+      OPENSSL_ROOT_DIR: C:/vcpkg/packages/openssl_x64-windows
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ilammy/msvc-dev-cmd@v1
```
absurdfarce, line 16: Appears to be necessary to setup MSVC env properly.  More at [the Githubs](https://github.com/ilammy/msvc-dev-cmd).

src/gssapi/dse_auth_gssapi.cpp
```diff
@@ -22,7 +22,6 @@
 #include <string.h>
 
 #include <gssapi/gssapi.h>
-#include <gssapi/gssapi_generic.h>
```
absurdfarce, line 25: This header file isn't included in the krb5 vcpkg build.  At least for modern versions of the gssapi lib this header isn't [especially interesting](https://github.com/krb5/krb5/blob/master/src/lib/gssapi/generic/gssapi_generic.h) so it seemed easy enough to remove.

absurdfarce:
Worth noting that this isn't 💯 yet.  Outstanding issues at this writing:

* Only running against Windows 2019, need to setup a matrix
* We're seeing several test failures due to missing host configs
* Newer versions of MSVC include checks for memory leaks and apparently some of our unit tests are triggering those

I'm not necessarily looking for bullet-proof answers to each of these points before we merge this; I expect this process will iteratively get better over time.

PR #568, opened 2024-11-11
Make cassandra_static build using position independent code

meghprkh:
This should help other toolchains.

If needed this can be gated behind a config flag.


meghprkh:
Please let me know what I can do to help with this ticket

PR #569, opened 2024-11-15
CPP-999 Compilation error in cpp-driver-2.16.2 when using C++98 standard

absurdfarce:
Remove C++11 usage in favor of a struct constructor to be explicit about the zero value init

### DIFF
--- a/src/metrics.hpp
+++ b/src/metrics.hpp
@@ -267,19 +267,6 @@ public:
       int64_t percentile_98th;
       int64_t percentile_99th;
       int64_t percentile_999th;
-
-      Snapshot() {
-        min = 0;
-        max = 0;
-        mean = 0;
-        stddev = 0;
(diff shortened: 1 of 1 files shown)


absurdfarce:
Don't overthink this; let's solve this the easy way via a base constructor for the snapshot struct.  Makes sense for that as a default anyway and you avoid any concerns about default val initialization.

absurdfarce:
Things to look into/fix after review with @lukasz-antoniak:

* Add the explicit call to zero arg constructor for zero_snapshot_ (just to be explicit)
* Can we create a singleton zero_snapshot_ rather than a per-histogram val?

absurdfarce:
Confirmed that re-compile with `--std=c++98` flag no longer complains about the initializer in this file (since there isn't an initializer anymore).

PR #575, opened 2025-09-25
CASSCPP-5 cass_cluster_set_load_balance_dc_aware_n() fails with NULL/empty local DC

absurdfarce:
Replacement for #571 

Description from original:
As a nice self-contained issue I thought this might be an interesting test case to see how Claude Code might handle this change. The basic change was identified pretty readily (which is impressive even in a constrained case like this) but the inconsistent behaviour between functions was a bit surprising to me.

### DIFF
--- a/src/cluster_config.cpp
+++ b/src/cluster_config.cpp
@@ -281,9 +281,6 @@ void cass_cluster_set_load_balance_round_robin(CassCluster* cluster) {
 CassError cass_cluster_set_load_balance_dc_aware(CassCluster* cluster, const char* local_dc,
                                                  unsigned used_hosts_per_remote_dc,
                                                  cass_bool_t allow_remote_dcs_for_local_cl) {
-  if (local_dc == NULL) {
-    return CASS_ERROR_LIB_BAD_PARAMS;
-  }
   return cass_cluster_set_load_balance_dc_aware_n(cluster, local_dc, SAFE_STRLEN(local_dc),
                                                   used_hosts_per_remote_dc,
                                                   allow_remote_dcs_for_local_cl);
@@ -293,11 +290,13 @@ CassError cass_cluster_set_load_balance_dc_aware_n(CassCluster* cluster, const c
                                                    size_t local_dc_length,
                                                    unsigned used_hosts_per_remote_dc,
                                                    cass_bool_t allow_remote_dcs_for_local_cl) {
-  if (local_dc == NULL || local_dc_length == 0) {
-    return CASS_ERROR_LIB_BAD_PARAMS;
-  }
+  // If local_dc is NULL or length is 0, we use an empty string which causes the driver
+  // to use the DC of the first connected node
+  String dc_name = (local_dc != NULL && local_dc_length > 0) ? 
+                    String(local_dc, local_dc_length) : String();
+  
   cluster->config().set_load_balancing_policy(new DCAwarePolicy(
-      String(local_dc, local_dc_length), used_hosts_per_remote_dc, !allow_remote_dcs_for_local_cl));
+      dc_name, used_hosts_per_remote_dc, !allow_remote_dcs_for_local_cl));
   return CASS_OK;
 }
 
--- a/tests/src/unit/tests/test_cluster_config.cpp
+++ b/tests/src/unit/tests/test_cluster_config.cpp
@@ -1,0 +1,160 @@
+/*
+  Copyright (c) DataStax, Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+#include <gtest/gtest.h>
+
+#include "cluster_config.hpp"
+#include "cassandra.h"
+#include "dc_aware_policy.hpp"
+#include "string.hpp"
+
+using namespace datastax;
+using namespace datastax::internal;
+using namespace datastax::internal::core;
+
+class ClusterConfigUnitTest : public testing::Test {
+public:
+  ClusterConfigUnitTest() {}
+
+  virtual void SetUp() {
+    cluster_ = cass_cluster_new();
+  }
+
+  virtual void TearDown() {
+    cass_cluster_free(cluster_);
+  }
+
+protected:
+  CassCluster* cluster_;
+
+  const DCAwarePolicy* build_dc_policy() {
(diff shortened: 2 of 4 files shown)


absurdfarce:
@yifan-c Another one for 2.17.2

absurdfarce:
Jenkins build looks good except for some weird failures with Rocky Linux 9.  These failures look to be the same as what was described [here](https://github.com/apache/cassandra-cpp-driver/pull/583#issuecomment-4049717924).  Seems like we've prolly got something borked in one of our Jenkins runners... I'll investigate that elsewhere, but for now it shouldn't hold up work on this ticket.

src/cluster_config.cpp
```diff
@@ -295,11 +296,13 @@ CassError cass_cluster_set_load_balance_dc_aware_n(CassCluster* cluster, const c
...
                                                    unsigned used_hosts_per_remote_dc,
                                                    cass_bool_t allow_remote_dcs_for_local_cl) {
-  if (local_dc == NULL || local_dc_length == 0) {
-    return CASS_ERROR_LIB_BAD_PARAMS;
-  }
+  // If local_dc is NULL or length is 0, we use an empty string which causes the driver
+  // to use the DC of the first connected node
+  String dc_name = (local_dc != NULL && local_dc_length > 0) ? 
+                    String(local_dc, local_dc_length) : String();
+
```
yifan-c: Since now `cass_cluster_set_load_balance_dc_aware_n` accepts NULL `local_dc`, can we remove the duplicated NULL check?

For example, 

```c++
CassError cass_cluster_set_load_balance_dc_aware(CassCluster* cluster, const char* local_dc,
                                                 unsigned used_hosts_per_remote_dc,
                                                 cass_bool_t allow_remote_dcs_for_local_cl) {
  return cass_cluster_set_load_balance_dc_aware_n(cluster, local_dc, SAFE_STRLEN(local_dc),
                                                  used_hosts_per_remote_dc,
                                                  allow_remote_dcs_for_local_cl);
}

CassError cass_cluster_set_load_balance_dc_aware_n(CassCluster* cluster, const char* local_dc,
                                                   size_t local_dc_length,
                                                   unsigned used_hosts_per_remote_dc,
                                                   cass_bool_t allow_remote_dcs_for_local_cl) {
  // NULL or zero-length local_dc means auto-detect from the first connected node
  cluster->config().set_load_balancing_policy(new DCAwarePolicy(
      String(local_dc != NULL && local_dc_length > 0 ? local_dc : "", local_dc_length),
      used_hosts_per_remote_dc, !allow_remote_dcs_for_local_cl));
  return CASS_OK;
}
```

tests/src/unit/tests/test_cluster_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+  EXPECT_EQ(CASS_OK, cass_cluster_set_load_balance_dc_aware(
+                         cluster_, "", 2, cass_true));
+
+  const DCAwarePolicy* dc_policy = build_dc_policy();
+
+  // Should be using empty string as local DC (will be determined at runtime)
+  EXPECT_EQ(dc_policy->local_dc(), String());
+  EXPECT_EQ(dc_policy->used_hosts_per_remote_dc(), 2u);
+  EXPECT_FALSE(dc_policy->skip_remote_dcs_for_local_cl());
+}
```
yifan-c: nit: new line at the end

tests/src/unit/tests/test_cluster_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+  virtual void SetUp() {
+    cluster_ = cass_cluster_new();
+  }
+
+  virtual void TearDown() {
+    cass_cluster_free(cluster_);
+  }
+
+private:
```
yifan-c: Is it redundant?

tests/src/unit/tests/test_cluster_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+  const DCAwarePolicy* dc_policy = build_dc_policy();
+
+  // Should be using empty string as local DC (will be determined at runtime)
+  EXPECT_EQ(dc_policy->local_dc(), String());
+  EXPECT_EQ(dc_policy->used_hosts_per_remote_dc(), 3u);
+  EXPECT_TRUE(dc_policy->skip_remote_dcs_for_local_cl());
+}
+  
+TEST_F(ClusterConfigUnitTest, SetLoadBalanceDcAwareWithEmptyLocalDc) {
+    // Test with empty string to use local DC from connected node
```
yifan-c: nit: 4 space indentation, should probably be 2 to be consistent with other comments.

tests/src/unit/tests/test_cluster_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+  size_t partial_length = 5; // Should just use "my_da" as the datacenter name
+  EXPECT_EQ(CASS_OK, cass_cluster_set_load_balance_dc_aware_n(
+                         cluster_, long_dc_name, partial_length, 
+                         2, cass_true));
+
+  const DCAwarePolicy* dc_policy = build_dc_policy();
+
+  // Should be using the partial string as local DC
+  EXPECT_EQ(dc_policy->local_dc(), String(long_dc_name, partial_length));
+  EXPECT_EQ(dc_policy->local_dc(), String("my_da"));
```
yifan-c: Are they identical?

tests/src/unit/tests/test_cluster_config.cpp
absurdfarce: Added

tests/src/unit/tests/test_cluster_config.cpp
absurdfarce: Looks to be.  I'm guessing Claude was being excessively explicit here.

tests/src/unit/tests/test_cluster_config.cpp
absurdfarce: Fixed.

tests/src/unit/tests/test_cluster_config.cpp
absurdfarce: They are.  I'm removing the check against the literal string.  Ideally we'd want this test to set that value once and refer to it when needed throughout the test; that allows us to change that test string and know the rest of the test should be fine.

Fixing.

src/cluster_config.cpp
absurdfarce: The duplicate null and empty checks do indeed look redundant.  I removed those but decided to keep the local dc_name var; I think it makes the resulting function easier to read.

Built and ran the unit tests locally with that change and everything passed so there's no obvious hidden bug waiting for us there.  I'll kick off a full Jenkins run once this change goes in to explicitly confirm.

absurdfarce:
Jenkins run for recent changes was successful.

absurdfarce:
@yifan-c can you take another look at this one and give me a +1/approval if you're satisfied with the current state?  I'm trying to put everything together to get 2.17.2 out the door and I'd really like to make sure this one is included.

yifan-c (approved):

PR #576, opened 2025-09-25
CPP-1006 Initial version of a Windows config for Github Actions

absurdfarce:
Replacement for #567


absurdfarce:
Pausing this effort for now.  Now that the driver has been donated to the ASF we need to think a bit more about whether something like this has a role going forward.

PR #578, opened 2025-09-25
CPP-953 Naive attempt to avoid pure virtual function calls at cluster shutdown

absurdfarce:
Replacement for #524


absurdfarce:
Closing this in favor of a better solution

PR #580, opened 2025-12-31
Bump minimum required version for cmake4 compatibility

Sigmanificient:
CMake 4, released half a year ago, [has dropped compatibility with versions older than 3.5](https://cmake.org/cmake/help/v4.0/release/4.0.html#deprecated-and-removed-features).

This patch bumps the minimal required version of CMake to 3.10.

I've set it to 3.10, so it doesn’t get deprecated as soon as 3.5 will be, and yet doesn’t impose a too new CMake release to build. For reference, [the oldest maintained LTS debian already has 3.18](https://packages.debian.org/bullseye/cmake).

This issue has come up on [nixpkgs](https://nixos.org), where a migration to CMake 4 is underway.

See also:
- NixOS/nixpkgs/issues/445447
- https://github.com/NixOS/nixpkgs/pull/475601

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.10)
 project(cassandra C CXX)
 
 set(CASS_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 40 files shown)


absurdfarce:
Thanks for the PR @Sigmanificient!  Apologies for the delay, I've been out for the holidays... I'll try to get you a review in the next few days!

absurdfarce:
This seems right to me.  I note for the record that [cmake 3.10.0](https://github.com/Kitware/CMake/releases/tag/v3.10.0) was released in late November of 2018 so we're not imposing an enormous burden to update to a brand new version most distros won't have.

absurdfarce (approved):

yifan-c (approved):

yifan-c:
Test it locally to build and it was successful.

PR #582, opened 2026-01-14
add support for appleclang and update cmake

gold-development:
Added support for AppleClang and update cmake version

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.5)
 project(cassandra C CXX)
 
 set(CASS_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
@@ -162,6 +162,7 @@ endif()
 #------------------------
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+   "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" OR
    "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
   # Enable C++11 support to use std::atomic
   if(CASS_USE_STD_ATOMIC)
@@ -177,7 +178,8 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
     endif()
   endif()
 
-  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") 
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+     "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang") 
     # Clang/Intel specific compiler options
     # I disabled long-long warning because boost generates about 50 such warnings
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic -Wextra -Wno-long-long -Wno-unused-parameter")
--- a/cmake/ClangFormat.cmake
+++ b/cmake/ClangFormat.cmake
@@ -1,7 +1,7 @@
 #
 # Format and verify formatting using clang-format
 #
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.5)
 
 include(FindPackageHandleStandardArgs)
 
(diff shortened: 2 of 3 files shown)


silviucpp:
Mac os build seems broken . any plan to commit this ?

absurdfarce:
Apologies, I've been distracted by too many things.  @yifan-c can you take a look at this one as well?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.5)
```
absurdfarce, line 1: Can you elaborate on a couple things here?

First: why is it necessary to change the cmake minimum version at all in order to fix this?

Second: if it _is_ necessary to change it why are we tweaking it only in three CMakeLists files rather than uniformly throughout the project?

CMakeLists.txt
```diff
@@ -177,7 +178,8 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
     endif()
   endif()
 
-  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") 
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+     "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
```
absurdfarce, line 182: I'll let him speak on this for sure but I believe @yifan-c was able to get this working more generally using regex matching:

```
if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang$" 
```

That's probably more desirable as it covers a broader range of clang derivatives.  cmake supports [a number of such variants](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html) so as long as the flags in question apply to clang generally (i.e. isn't some custom flag implemented for a specific clang variant) the more general syntax seems preferrable.

absurdfarce (commented):
A few questions but I'll defer to @yifan-c on the details since I know he's been working much more directly in this space.

absurdfarce:
@gold-development Are you able to speak to any of the comments on this PR?  I'd love to get this change into 2.17.2 but I do think we have a few points to talk about in the changes above.

CMakeLists.txt
toptobes: Looking at the git logs, I'm fairly certain what happened is that they were originally basing this off the version of the c++ driver 2 weeks before this PR was created, when it was still using cmake version `2.8.12` (so this was intended as an upgrade, not a downgrade)

See:
```rb
❯ git lg                               
* a22a11a3 golddevelopment - (HEAD -> gold-development/trunk) add support for appleclang and update cmake (5 months ago)
* e5a486ac Sigmanificient - Bump minimum required version for cmake4 compatibility (5 months ago)
```

```rb
❯ git lg -L 1,1:CMakeLists.txt                    
* a22a11a3 golddevelopment - (HEAD -> gold-development/trunk) add support for appleclang and update cmake (5 months ago)| 
| diff --git a/CMakeLists.txt b/CMakeLists.txt
| --- a/CMakeLists.txt
| +++ b/CMakeLists.txt
| @@ -1,1 +1,1 @@
| -cmake_minimum_required(VERSION 3.10)
| +cmake_minimum_required(VERSION 3.5)

* e5a486ac Sigmanificient - Bump minimum required version for cmake4 compatibility (5 months ago)| 
| diff --git a/CMakeLists.txt b/CMakeLists.txt
| --- a/CMakeLists.txt
| +++ b/CMakeLists.txt
| @@ -1,1 +1,1 @@
| -cmake_minimum_required(VERSION 2.8.12)
| +cmake_minimum_required(VERSION 3.10)
```

[P.S. `AppleClang` only got differentiated from `Clang` in `3.0` which is why this wasn't an issue previously](https://cmake.org/cmake/help/latest/policy/CMP0025.html)

absurdfarce: This seems reasonable @toptobes but even if it is the case I'd still argue for a uniform cmake requirement throughout the project.  Note [the commit](https://github.com/apache/cassandra-cpp-driver/commit/e5a486ac52b99ae74d40e285b4438252fcf3b07a) referenced above; in that case we updated cmake_minimum_required() throughout the project.  If we're going to change the cmake requirement _at all_ here it seems like we should do the same.

toptobes: ➕

absurdfarce:
Closing in favor of #586.  If there's some part of this you'd like to revisit in the future @gold-development please feel free to open a new PR!

PR #583, opened 2026-03-12
CASSCPP-3 Crash on Pure virtual function while the destructor of the Session object is called

absurdfarce:
Multiple steps to try to keep us out of the crash situation.  First: remove the call to the callback from the handler's on_close() function.  It's not especially useful when a connection is in the process of closing down like this; when that happens host up or host add events (the objective of the callback in question) should be considered unreliable anyway.  This change alone is _probably_ enough to fix the problem but it's effect there is entirely one of timing; by not having to handle the extra event delivery logic we're not taking as long in the callback so there's less opportunity to overlap with other resources being destructed.  That said, minimizing the amount of work here is still a pretty good idea.

Second: don't make the on_close() function of the handler a pure virtual function.  There's no obvious reason this function has to be a _pure_ virtual function so the hope is that by giving it a (no-op) body we'll just execute that if we wind up with a conversion from a PrepareHostHandler to a ConnectionListener object while the event loop is shutting down.

### DIFF
--- a/src/connection.hpp
+++ b/src/connection.hpp
@@ -87,7 +87,7 @@ public:
    *
    * @param connection The closing connection.
    */
-  virtual void on_close(Connection* connection) = 0;
+  virtual void on_close(Connection* connection) {}
 };
 
 /**
--- a/src/prepare_host_handler.cpp
+++ b/src/prepare_host_handler.cpp
@@ -69,8 +69,6 @@ void PrepareHostHandler::prepare(uv_loop_t* loop, const ConnectionSettings& sett
 }
 
 void PrepareHostHandler::on_close(Connection* connection) {
-  callback_(this);
-
   dec_ref(); // The event loop is done with this handler
 }
 
--- a/src/cluster.cpp
+++ b/src/cluster.cpp
@@ -663,6 +663,19 @@ void Cluster::notify_or_record(const ClusterEvent& event) {
   }
 }
 
+// TODO: This function takes a callback because it's called from both the host up and host add handlers.
+// The usage there is to send appropriate notifications to all other nodes whether query prep succeeds or
+// not.  But that processing isn't really part of what _this_ function is trying to accomplish; that
+// responsibility really relies with the host up or host add handler itself.
+//
+// A more robust implementation would be to implement the host up and host add handlers as a combined
+// sequence of ops and then wait on that whole sequence to either complete or fail.  That's a fairly
+// significant change to the current impl, though, so for now just moving the notifications into
+// the host up/add handlers will be sufficient.
+//
+// Why does any of this matter?  Originally we were triggering this callback on socket close even when
+// we were closing a connection.  This link has already been severed (see CASSCPP-3) but that effort
+// exposed the unnecessary entanglement described above.
 bool Cluster::prepare_host(const Host::Ptr& host, const PrepareHostHandler::Callback& callback) {
   if (connection_ && settings_.prepare_on_up_or_add_host) {
     PrepareHostHandler::Ptr prepare_host_handler(


src/prepare_host_handler.cpp
```diff
@@ -69,8 +69,6 @@ void PrepareHostHandler::prepare(uv_loop_t* loop, const ConnectionSettings& sett
 }
 
 void PrepareHostHandler::on_close(Connection* connection) {
-  callback_(this);
```
absurdfarce, line 72: Long-term the better answer is probably to not make HOST_UP/HOST_READY/HOST_ADD event distribution a function of the callback in use here.  Those ops should be handled by logic in the relevant functions of cluster.cpp, specifically as some kind of "after" action from the prepare op.  But this fix allows us to address the immediate problem without having to refactor the entirety of message delivery.

absurdfarce:
@yifan-c If you have a sec I'd love a review of this one!

absurdfarce:
IBM Jenkins run for this change was clean.  Well, there were some weird failures around cloud connectivity for Rocky Linux 9 only but I'm quite sure those were environmental/runner issues and not anything related to this change.

src/connection.hpp
```diff
@@ -87,7 +87,7 @@ class ConnectionListener {
    *
    * @param connection The closing connection.
    */
-  virtual void on_close(Connection* connection) = 0;
+  virtual void on_close(Connection* connection) {};
```
yifan-c: the tailing `;` is redundant.

src/prepare_host_handler.cpp
yifan-c: Thanks for sharing the context. Shall we comment the `todo` in the code?

src/connection.hpp
absurdfarce: Good catch, fixing

src/prepare_host_handler.cpp
absurdfarce: Agreed.  I've added a TODO in the code.

I'm not sure it needs an actual _ticket_ yet; if feels like a fairly big refactoring to take on and the change in this PR should sever the problematic link anyway so we _should_ have a short-term fix now.

absurdfarce:
Jenkins run for recent changes was successful.

ankur-dauneria:
Hi, any update on when version 2.17.2 is planned to be released? Thanks!

absurdfarce:
Apologies for the delay @ankur-dauneria... trying to juggle too many things.

2.17.2 has definitely taken longer to get out than I'd planned on.  I'm considering shrinking up this release to include the fixes in review so far (including this fix) and [CASSCPP-6](https://issues.apache.org/jira/browse/CASSCPP-6), mainly to facilitate getting our first ASF CPP release out more quickly.  We'd then move on to a 2.18.0 which would include the other things originally slated for 2.17.2 as well as a few other (more significant) fixes.

If we pursue that course I would expect 2.17.2 to come together fairly quickly.  We have PRs for two out of the three fixes I want to include and I have a good idea for how to address CASSCPP-6.  It just... needs to get done.

absurdfarce:
Same with this one @yifan-c.  I'd like for this change to be included in 2.17.2 as well but to make that happen I'd really like an explicit +1 from you saying that the current state of this change looks good.

ankur-dauneria:
> If we pursue that course I would expect 2.17.2 to come together fairly quickly. We have PRs for two out

Thanks for the update! 

One question  - Is the 2.17.2 release expected in 2026, or could it slip to 2027 ?

yifan-c (approved):
Sorry for the delay. I am +1

absurdfarce:
Hey @ankur-dauneria, my expectation is that 2.17.2 will be released in the next couple weeks.  We still have a few tickets to finish up and there are a few questions about what distribution looks like but my expectation is that those will get sorted out sooner rather than later.

ankur-dauneria:
> Hey @ankur-dauneria, my expectation is that 2.17.2 will be released in the next couple weeks. We still have a few tickets to finish up and there are a few questions about what distribution looks like but my expectation is that those will get sorted out sooner rather than later.

Thanks!

ankur-dauneria:
> Hey @ankur-dauneria, my expectation is that 2.17.2 will be released in the next couple weeks. We still have a few tickets to finish up and there are a few questions about what distribution looks like but my expectation is that those will get sorted out sooner rather than later.

 Hey @absurdfarce , just checking in on 2.17.2 release—any blockers holding it up, or has the timeline shifted?

absurdfarce:
Thanks for checking @ankur-dauneria.  This release has been pre-empted by work on the Python driver 3.30.1 and a Java driver release for the 3.x series.  The Python driver release is [out now](https://pypi.org/project/cassandra-driver/3.30.1/) but work on the 3.x Java driver is ongoing.  I'm hoping to get [CASSCPP-6](https://issues.apache.org/jira/browse/CASSCPP-6) in soon (that's really the last holdout) and then push 2.17.2 out the door.

absurdfarce:
If you'd like to follow along more closely and/or engage in the discussion around this driver I'm happy to get you an invite to [ASF Slack](https://infra.apache.org/slack.html) (where most of these discussions take place).  If you're interested just let me know and I'll set you up!

